PR #6, opened 2020-01-08
Add for-await support for streaming

### LINKED ISSUE #1: Stream speed management using async/await (cursor)
The current stream mechanics make it easy to process rows as they are read from the query rather than return them in a block; however in ultra-large result set cases it's possible the rows will be read faster than they can be processed.  See this [Reddit thread](https://www.reddit.com/r/PostgreSQL/comments/ef2zwc/github_porsagerpostgres_fastest_full_featured/fbybsoi/) for more comprehensive specifics.  An example use case would be executing a query that yields 1 million+ rows then sending them in blocks of 200 to an external web service that takes several seconds to process each request.  In this case it could be possible to send hundreds or thousands of concurrent outbound requests overwhelming heap space and the external service.

There are two potential fixes:

1.  Allow iterator functions to return a promise which is then awaited to ensure more data is not sent to the iterator function than it can handle.  The following is a hypothetical use case that buffers up to 200 rows at a time then sends it to an external service with limited speed:
```javascript
	var rowBuffer = [];
	await sql`select created_at, name from events`.stream(async row => {
		if (rowBuffer.push(row) >= 200) await transformAndSend(rowBuffer.splice(0,200));
	});
	if (rowBuffer.length > 0) await transformAndSend(rowBuffer.splice(0,200));
```
From a syntax standpoint all that would change from existing functionality is the addition of the optional async.  But from the backend the return from the result of the iterator would need to be checked if it is a promise, then awaited.  This is an overtly simplified example:
```javascript
	while (moreRows) {
		var row = await getRow();
		var P = StreamIterator(row);
		if (P instanceof Promise) await P;
	}
```

I'm not sure how the rows are received weather each is electively read or they come in as events which can be paused.  If they are purposefully read then this would be pretty easy.  You'd just await the callback and get the next row.  If they come in as events you'd need to buffer them to be sent to the callback and pause the incoming results if that buffer gets too large.I

\*Ideally the promise resolving to a true or a specified enum should stop the query from completing.  So if for example the user executed a query that botches a join and results in 10 billion rows of nonsense or the designation service won't accept the results in the callback, it would be nice to have a means to gracefully and forcefully stop it from reading more rows.

2. Alternatively (or additionally) making use of [Symbol.asyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) would allow a standard "for await" loop as a means to asynchronously process data from an async source.  This would be (very) easy to add in after altering to existing stream functionality to watch for an await returned promises and could be officially extended from the query result as an .iterate() that returns an Symbol.asyncIterator object to manage this flow for the end user.  That would look something like the following in practice:
```javascript
	var rowBuffer = [];
	for await (let row of sql`select created_at, name from events`.iterate()) {
		if (rowBuffer.push(row) >= 200) await transformAndSend(rowBuffer.splice(0,200));
	});
	if (rowBuffer.length > 0) await transformAndSend(rowBuffer.splice(0,200));
```
I'm at your disposal if you need assistance in implementing this feature or need to test it once complete.  If you can make efficiently iterating through large queries as easy as a "for await" loop you'll have introduced an quality of life improvement well ahead of the curve in terms of future proofing.

Minigugus:
Fix #1

This PR is a very simple and naive implementation of `for-await` for consuming stream. I added tests and updated the README.

Feel free to update or close if you have a better idea.

Otherwise, awesome project, thank you :smile: 

```js
let result = 0
for await (const { generate_series: x } of sql`select * from generate_series(1, 100000)`)
  result += x;
return [4699408878, result]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,9 @@ const users = await sql`
 
 ```
 
-## Stream ```sql` `.stream(fn) -> Promise```
+## Streams
+
+### ```sql` `.stream(fn) -> Promise```
 
 If you want to handle rows returned by a query one by one you can use `.stream` which returns a promise that resolves once there are no more rows.
 ```js
@@ -116,6 +118,21 @@ await sql`
 `.stream(row => {
   // row = { created_at: '2019-11-22T14:22:00Z', name: 'connected' }
 })
+
+// No more rows
+
+```
+
+### With `for-await`
+
+If you are using a recent version of Node.JS (> v10), can use the new `for-await` to consume streams properly.
+```js
+
+for await (const row of await sql`
+  select created_at, name from events
+`) {
+  // row = { created_at: '2019-11-22T14:22:00Z', name: 'connected' }
+}
 
 // No more rows
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -169,6 +169,25 @@ function Postgres(url, options) {
     })
 
     promise.stream = (fn) => (query.stream = fn, promise)
+    
+    if (Symbol.asyncIterator)
+      promise[Symbol.asyncIterator] = () => {
+        let locked = [], resolved = [];
+        query.stream = row => locked.length ? locked.pop()({ done: false, value: row }) : resolved.push(Promise.resolve({ done: false, value: row }));
+        promise.then(() => locked.length ? locked.pop()({ done: true }) : resolved.push(Promise.resolve({ done: true })))
+        return {
+          return() {
+            // TODO Abort the query properly
+            return Promise.resolve({ done: true });
+          },
+          next() {
+            return resolved.length ? resolved.pop() : new Promise(res => locked.push(res)); 
+          },
+          throw() {
+            return Promise.resolve({ done: true });
+          }
+        }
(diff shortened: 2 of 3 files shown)


porsager:
Hey @Minigugus ... Thanks a lot for the kind words and the PR.

As you might have seen in the #1 issue I've already got a working branch here: https://github.com/porsager/postgres/tree/cursor which I realize I should have made a PR for so that it was visible work was already under way.

With regards to your implementation here, while it works, it will fill up memory with all the rows coming from the database which I don't feel is what a user will expect if using `for await`. 

I'll be finishing up my work in the cursor branch this or next week which will also include `for await` support, so I hope you're ok with me closing out this PR?

Minigugus:
Argh, too fast, my bad :sweat_smile: 

Your implementation will obviously be better than mine, I'm closing this PR :wink:

Keep up the great work :smile:

porsager:
Cool :) 

And thanks, I'll do my best.

PR #8, opened 2020-01-09
Typescript support

### LINKED ISSUE #7: Typescript support
What about supporting Typescript ?
Just a simple `types.d.ts` would be enough, no need to fully rewrite the project :smile: 

Pro
 * Much more easy than reading the doc
 * Direct documentation with JSdoc and parameters validation (intellisense)

Con
 * Must be in sync with the implemented code

Minigugus:
Fix #7 :smile: 

Typescript definitions infered from the code and the documentation.

I relied on `tests/index.js` to improve theses definitions.
Just add `// @ts-check` at the beginning to enable type validation in js files.


porsager:
Awesome - that was fast!

It looks great, and very nice you were able to infer so much from the tests and docs. 
I'll try to go through it and test it, but I probably won't have time until the weekend.

What did you have in mind wrt. JSDoc?

Minigugus:
For JSdoc, I mean adding text documentation directly on types definitions to display it directly when the user is typing. [Here is an example](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=27&pc=1#code/JYOwLgpgTgZghgYwgAgIoFdoE8AKUD2AtsAM4oQAekIAJicnkaRADxwhYDaAugHzIBvAFDJRyAPQAqSSLHJJyAIIIkJegQDu9UMjjISYKBDjEQAc2SF2IaABpd9MAAsIWXVCjAAbhAB0ssQUAFSdSZBh0EAQwYHwQZBo4gHIwSzgAaxR0MmR8GGQAYXQoEnwoewAHIxhoZAADBGLSqDrwsuQAI2ALAEdMTwgSfzl5ZAABCrgoE2QEDsK4ABtFjsR05FAvfEyaNqhkYwQnZE1kDRd4uC84YEW4DsWIANEFMaMwYpB6RWQqphyjKVFl5QBZzhBLssTvgtO4UIZjJAaM95OJngYjCYABRzABcyCxmnxAmQnAQ+EW6EIIHxGNB3Hx7DcAF8AJTIAC8-C2wBorPxjGIZDYHB4vAA3EJnlIZCMFMpVOoYdpLvoESZQWkQDZyg5kM5XO5PD57NlNQBlVAAGUKTTKQxRCgAqmQIos9sg+tBgIMzsBnLpkIt8Kk8tCtMM5K9JtNCLN5gUlis1hsQFsdh7DsdTuDLtdbvdHo7xu9Pt9fgQhShARSQeYzhddFDNPQpvDMUjHWi5I0SmUcU5IulzcAAF4QfEgKkdOzx-GEmHE0nkynU2mGemMjjINmc7n4Xn8hiV5girh8SXMqU0CAIO5GZDXfYkHqLeeQQgVO6QfFBCCf78IHNDdzBIRQPDgLB7F8GCpjMEgt3PdkuTQfpcBPMhJSELE4BILAogJZD+GEHs4gMT00M5fRXzqc0AFErTogogmQABGOosLkcRxEKIgvwgSB9RcWZ8BvToIGDDRpR4r0oCwfw2SxVlJSAA).

Many packages do not release JSdoc with types definitions. That's a shame because it's really useful and saves a lot of time for beginners :smile: 

Can also be useful to depreciate some functions when the code evolves or to explain the user when a function should be used or not. It's also possible to add examples (that's what [`sequelize`](https://www.npmjs.com/package/sequelize) does).

porsager:
Oh it's really nice they're on the types, and not the actual code. I'd like to keep the code clean, so a solution like this seems really nice!

KristianFJones:
I need this module in one of my projects. I'm fairly good at creating conditional and dynamic types, anything I can help out with on this PR?

porsager:
Yeah, if you could check out what @Minigugus already did, and give feedback if any, that'd be really nice.

TobiasNickel:
actually, I also already started to develop a type definition, but yours, @Minigugus, is so much better then what I came up with.

Do you think we could support a syntax like the following:
```ts
interface INumberResult {
  n: number
}
const result = await sql<INumberResult>`select 1 as n;`
```
now when typing `result[0].` the IDE can offer `n` for autocomplete. 
?

Minigugus:
@TobiasNickel As explained [here](https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes), passing function return types as a generic parameter is considered a common mistake, such as `getMeAT<T>(): T` which is simply `getMeAT() as T` when called. Also, there are no constraints to ensure that the type parameter correctly reflect the query, and this is the main reason I decided not to include this feature.

However, after some experimentation, your proposal seems to be a solution for multi-statements queries that can't be properly typed since there is no way to detect whether the query is a single statement or multi statements query. Your proposal allows users to pass their own array when using multi-statements queries :wink: 

So, as you can see, I just added this feature, which means :
```typescript
interface INumberResult {
  n: number;
}

const a = await sql`select 1 as n`; // does not enforce the result type (any by default)
a[0].n; // ok
a[0].a; // ok
a[1].n; // ok

const b = await sql<INumberResult>`select 1 as n`;
b[0].n; // ok
b[0].a; // fails
b[1].n; // ok

const c = await sql<INumberResult[]>`select 1 as n`; // same as above
c[0].n; // ok
c[0].a; // fails
c[1].n; // ok

const d = await sql<[INumberResult]>`select 1 as m`; // errors in queries not detected
d[0].n; // ok
d[0].a; // fails
d[1].n; // ok (but should fail https://stackoverflow.com/a/52490977)
d.count; // shows 1 in the editor (the length of the tuple)

const e = await sql<[[INumberResult], [INumberResult]]>`select 1 as n; select 2 as n;`; // multi statements
e[0].n; // fails
e[0][0].n; // ok (access the first column of the first statement)
e[1][0].n; // ok (access the first column of the second statement) 
e[1][2].n; // fails (access the third column of the second statement)

const f = await sql.file<[[INumberResult], [INumberResult]]>('input.sql'); // works wherever a query is submitted
```

Can also be achieved without type parameters but it's a bit more verbose :
```typescript
const d: postgres.QueryResultArray<[INumberResult]> = sql`select 1 as n`;
```

Hope you enjoy :wink: 

> PS: You might detect some mistakes in English because it is not my primary language, please feel free to correct me or to participate! :smile:

porsager:
Just a note @Minigugus .. Multi statement queries are only possible when using `.unsafe()` and `.file()` without arguments..

porsager:
@TobiasNickel @Minigugus sounds like a good idea to get a first version released. I'll take some time to review tonight..

Minigugus:
@porsager For multi-statements queries, I can't reject user types like `INumberResult[][]` for calls as template tag, it's up to the user to pass a correct type.

Regarding the release, maybe you could add JSDoc at least for query helpers, as you know better how they work ? Anyway, the current version seems to correctly reflect the surface API, even if some types depend on the postgresql server and are therefore not complete (eg. [`ConnectionVariables`](https://github.com/porsager/postgres/blob/520a38d1dde453329d6038898cd242fdd7775441/types/index.d.ts#L81-L86)). :wink:

Minigugus:
I tried to add support for errors type (#17) but since they don't have their own class, the type definitions are quite weak and require unsafe casts:
```typescript
try {
  await sql`select wat`;
} catch (err) { // `err` is `any`
  const error: PostgresError = err; // cannot use `instanceof` as `PostgresError` is virtual (doesn't exists in javascript)
  switch (error.code) {
    case 'SASL_SIGNATURE_MISMATCH':
    case 'AUTH_TYPE_NOT_IMPLEMENTED': // `error.type` is available
    case 'MESSAGE_NOT_SUPPORTED':
      throw new Error('Server not supported');

    case 'CONNECTION_DESTROYED':
    case 'CONNECTION_CLOSED':
    case 'CONNECTION_ENDED':
      throw new Error(`Disconnected from ${error.port === undefined ? `${error.address}` : `${error.address}:${error.port}`}`);

    case 'MAX_PARAMETERS_EXCEEDED':
      throw new Error('Too many parameters');

    case 'NOT_TAGGED_CALL':
      throw new SyntaxError('Use the template string tag');
    
    default: // no reliable way to detect if it's not an SQLError
      const sqlError: SQLError = error; // `error` need be manually casted to `SQLError` because adding `SQLError` to the union `PostgresError` would break the completion and the validation of this example
      throw new Error(`[${sqlError.severity}] Error n°${sqlError.code} (${sqlError.name}) in ${sqlError.file}:${sqlError.line} : ${sqlError.message}`);
  }
}
```
So I'm wondering if this feature should be pushed :thinking: The best would be to add errors classes in the javascript code :
```typescript
try {
  await sql`select wat`;
} catch (err) { // `err` is `any`
  if (err instanceof GenericError || err instanceof AuthTypeNotImplementedError)
    switch (err.code) {
      case 'MAX_PARAMETERS_EXCEEDED':
        throw new Error('Too many parameters');
      case 'NOT_TAGGED_CALL':
        throw new SyntaxError('Use the template string tag');

      default: // MAX_PARAMETERS_EXCEEDED | SASL_SIGNATURE_MISMATCH | AUTH_TYPE_NOT_IMPLEMENTED
        throw new Error('Server not supported');
    }
  else if (err instanceof ConnectionError)
    throw new Error(`Disconnected from ${err.port === undefined ? `${err.address}` : `${err.address}:${err.port}`}`);
  else if (err instanceof SQLError)
    switch (err.code) {
      case '42703': // A column doesn't exists
        // Handle the error correctly
        break;

      default:
        throw new Error(`[${err.severity}] Error n°${err.code} (${err.name}) in ${err.file}:${err.line} : ${err.message}`);
    }
  else // `err` is something else
    throw err;
}
```
but it's beyond the scope of this pull request.

@porsager What do you think?

porsager:
I think that's a good idea @Minigugus. I'll get that done, and then expose the constructors on eg. `postgres.errors` ?

porsager:
I'm sorry I haven't had time to go through it yet @Minigugus . Even though I'm not into typescript, I probably have an opinion on naming 😝 I hope I'll get to it one of the nights this week ;)

Minigugus:
> I think that's a good idea @Minigugus. I'll get that done, and then expose the constructors on eg. `postgres.errors` ?

Yes, or directly on the main export for ES modules users : `import { AnError } from 'postgres';` :smile: 

> I'm sorry I haven't had time to go through it yet @Minigugus . Even though I'm not into typescript, I probably have an opinion on naming  I hope I'll get to it one of the nights this week ;)

Names are merely illustrative, feel free to change them :wink:

tamino-martinius:
To include the typings in the released package please add the `/types` folder also to the `"files"` property in package.json

types/index.d.ts
```diff
@@ -0,0 +1,215 @@
...
+  interface Notice {
+    [field: string]: string;
+  }
+
+  interface CustomType {
+    to: number,
+    from: number[],
+    serialize(obj: unknown): unknown,
+    parse(raw: any): unknown
+  }
```
davojan: Wouldn't type-parameter make more sense here?
```typescript
  interface CustomType<T> {
    to: number
    from: number[]
    serialize(obj: T): string
    parse(raw: string): T
  }
```

types/index.d.ts
```diff
@@ -0,0 +1,215 @@
...
+    // cursor(size: number): AsyncIterable<QueryResult>;
+    // cursor(size: number, cb: (row: QueryResult) => void): QueryResultPromise;
+  }
+
+  interface QueryParameter<T, U extends any[] = T[]> {
+    first: T;
+    rest: U;
+  }
+
+  interface Tag<TTypes extends CustomTypeList> {
```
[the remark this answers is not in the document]
Minigugus: `TTypes` refer to custom types passed when constructing the `sql` function :
```js
const sql = postgres({
  types: {
    rect: {
      to        : 1337,
      from      : [1337],
      serialize : ({ x, y, width, height }) => [x, y, width, height],
      parse     : ([x, y, width, height]) => { x, y, width, height }
    }
  }
});

sql.types.rect({ x: 13, y: 37: width: 42, height: 80 }); // ok, returns `number[]`
sql.types.square({ x: 13, y: 37: width: 42, height: 80 }); // fails (`square` not defined)
```

>Would it make sense if this was named Sql to match the naming used in the documentation?

Sure, I get it :wink:

types/index.d.ts
Minigugus: Not really, as the `T` type-parameter would have to be passed everywhere it is used, and moreover multiple types can be specified in custom types list, so a commun `T` for all is weak.

Instead of using a type-parameter, I let TypeScript infer the whole custom types list https://github.com/porsager/postgres/blob/4e5427086cea596de5a640813b1a95402fa727e2/types/index.d.ts#L133-L135
so that I can use it as a mapping type : https://github.com/porsager/postgres/blob/4e5427086cea596de5a640813b1a95402fa727e2/types/index.d.ts#L202-L204

porsager: It's the definition of how to handle specific postgres types, so maybe simply `PostgresType`?

types/index.d.ts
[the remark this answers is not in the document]
porsager: Ah, cool :) That makes sense..

types/index.d.ts
Minigugus: Should I rename `TTypes`, `CustomType` or `CustomTypeList` ?

porsager:
@Minigugus I'm ok with merging, but I'm also finishing up .cursor support and making those errors as we talked about, so would you rather want to wait and include that in this PR when I'm done, or we could look at those things later in a separate PR?

types/index.d.ts
porsager: No I think it's fine if you feel good about them, but if you think `PostgresType`, `PostgresTypeList` make better sense then feel free :)

I'm not sure about `TTypes`. If I understand correctly it's not something a typescript user would be exposed to, so it's not that important?

types/index.d.ts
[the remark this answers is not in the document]
davojan: > What does `TTypes` refer to / mean?

It refers to the `T` param from [this line](https://github.com/porsager/postgres/pull/8/files/4e5427086cea596de5a640813b1a95402fa727e2#diff-88bec6beae84369c0376b56b3bb88fe1R8):
```typescript
declare function Postgres<T extends Postgres.CustomTypeList = {}>(options?: Postgres.Options<T>): Postgres.Tag<T>
```
and allows to map `serialize` function types from the custom type declarations to the query helpers (`sql.types.rect` example in README) [here](https://github.com/porsager/postgres/pull/8/files/4e5427086cea596de5a640813b1a95402fa727e2#diff-88bec6beae84369c0376b56b3bb88fe1R203) (piece of typescript art 🙂)

Minigugus:
Adding cursor declaration is very simple (1 line), error declarations a bit longer (~10 to 20 lines) I guess. Maybe we could wait ? :wink:

porsager:
Cool.. I'll get on with that then ;)

types/index.d.ts
Minigugus: Normally I would have called it `T` but it is already used by functions in `Sql`, so I chose `TTypes`. Indeed, It is not exposed but can help to understand the file.

davojan: > Instead of using a type-parameter, I let TypeScript infer the whole custom types list

Yes, I got your intention. But may be we could push `unknown` a little bit higher:
```typescript
 interface CustomTypeList { 
   [name: string]: CustomType<unknown, unknown>
 }
```
and the compiler will at least type-check that `serialize` and `parse` functions does match each other? Not sure about that though...

karlhorky:
@porsager is there anything still open here before this is merged?

Minigugus:
@karlhorky https://github.com/porsager/postgres/pull/8#issuecomment-581632508 (would also be nice to have JSDoc but I'm not good enough in English :rofl: - help wanted :smile:)

karlhorky:
@Minigugus so which parts of this comment are still open? Both of these things below?

> adding cursor declaration is very simple (1 line)

> error declarations a bit longer (~10 to 20 lines) I guess

I thought that some of your latest commits in this branch addressed these things already...

Minigugus:
@karlhorky 
> I thought that some of your latest commits in this branch addressed these things already...

For the cusor only. I am waiting an explicit and accessible error list in JS.

karlhorky:
> I am waiting an explicit and accessible error list in JS

Oh, ok.

@porsager if you're not going to get to this error list soon, what do you think about merging this one as is now and adding the errors later in a separate PR?

porsager:
@karlhorky I've got started yesterday, and should have something today or tomorrow, so let's include it ;)

@Minigugus I've decided to go with a more simple setup which should be possible to type well, but better in practice (no ambigious grouping). I'll push a branch today or tomorrow for us to look at where I'd like your feedback too :)

types/index.d.ts
```diff
@@ -0,0 +1,166 @@
...
+    [name: string]: any;
+  }
+  
+  interface Options<T extends CustomTypeList> extends BaseOptions<T> {
+    /** unix socket path (usually '/tmp') */
+    path?: string;
+    /** Password of database user */
+    pass?: string | (() => string | Promise<string>);
+    /** Password of database user */
+    password?: Options<T>['pass']; // FIXME Is it a doc error ?
```
kukawski, line 92: Looking at the library code, both are valid and `pass` is checked before `password`.
There are more options that have "aliases", e.g. `hostname` and `host`, `database` and `db`, `user` and `username`. All these are combined in ParsedOptions into `host`, `database`, `user`, `pass`. In such case, should these be reflected as well in the typings?

types/index.d.ts
```diff
@@ -1,8 +1,22 @@
 /// <reference types="node" />
 
+/**
+ * Etablish a connection to a PostgreSQL server.
```
kukawski, line 2: Wording change suggestion: `Etablish` -> `Establish`
I'm not native English speaker, but usually the word establish is used regarding database connections, also many dictionaries don't include the word etablish and suggest establish.

porsager:
Hey @Minigugus ..

Sorry this took so long.. I almost had it ready, but didn't until now. I've pushed the error type changes here now: https://github.com/porsager/postgres/commit/fc6804a590eca3184286d02f1443e2bbcc31edae

Basically there is now a `PostgresError` type on sql.PostgresError now, and other checks can be made on `error.code`, but I don't know if you intended to type all those as an enum or something like it?.

porsager:
I have two last things I want to add before getting ready for a v2 release, but they shouldn't affect types, so if the Types could be updated to match all the latest changes to master we could get them merged for the v2 release. That would be really neat!

types/index.d.ts
Minigugus: I mostly infered the typings with the `tests/index.js` file, so it's possible some part was forgotten :smile: 
Thanks for the help :wink:

types/index.d.ts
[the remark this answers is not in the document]
Minigugus: Such precision! Tanks :laughing:

types/index.d.ts
```diff
@@ -1,8 +1,22 @@
 /// <reference types="node" />
 
+/**
+ * Etablish a connection to a PostgreSQL server. 
+ * @param options Connection options - default to the same as psql
+ * @returns An utility function to make queries to the server
+ */
 declare function Postgres<T extends Postgres.CustomTypeList = {}>(options?: Postgres.Options<T>): Postgres.Tag<T>
+/**
+ * Etablish a connection to a PostgreSQL server.
```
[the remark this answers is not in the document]
Minigugus: Thanks again :smiley:

types/index.d.ts
Minigugus: I'm not a native English speaker either, proof! :laughing:

Minigugus:
> Basically there is now a `PostgresError` type on sql.PostgresError now, and other checks can be made on `error.code`, but I don't know if you intended to type all those as an enum or something like it?.

It's perfect! TypeScript is going to love :laughing: I think a union type seems the way to go :wink: 

> I have two last things I want to add before getting ready for a v2 release, but they shouldn't affect types, so if the Types could be updated to match all the latest changes to master we could get them merged for the v2 release. That would be really neat!

That's what I planned to do :wink:

porsager:
Awesome @Minigugus !!

types/index.d.ts
```diff
@@ -0,0 +1,272 @@
...
+   */
+  function toCamel(str: string): string;
+  /**
+   * Convert a string to Kebab case.
+   * @param str THe string to convert
+   * @returns The new string in Kebab case
+   */
+  function toKebab(str: string): string;
+
+  const BitInt: PostgresType;
```
porsager, line 95: Should be BigInt ;)

Minigugus: Oups x)

cfanoulis:
Heyo! What's the status on this? Anything left that I could help with? I stumbled upon this and it's 10 times better than `node-postgres`, and would like to help with bringing this to TS`

Minigugus:
@cfanoulis Sure! You can help with the JSdoc (example below) :wink:
https://github.com/porsager/postgres/blob/0db879a79c3a445b2b67d57f97370c711dd4010c/types/index.d.ts#L268-L274

I know it's not really TS but it's the only thing that's missing and I'm not good enough in English to write it :cry: 

You can also review this code by checking if the typings are corrects, since I tested most of them only with the [`tests/index.js`](https://github.com/porsager/postgres/blob/0db879a79c3a445b2b67d57f97370c711dd4010c/tests/index.js) file :sweat_smile:

Minigugus:
Just a note regarding [shemats](https://github.com/SweetIQ/schemats#writing-code-with-typed-schema), a very useful project combined with typings proposed in this PR :

```ts
import * as osm from './osm'

// Now query with postgres and have a completely typed return value
let usersCreatedAfter2013
  = await sql<osm.users>`SELECT * FROM users WHERE creation_time >= '2013-01-01'`;

// We can decide to only get selected fields
let emailOfUsersCreatedAfter2013 = await sql<{
  email: osm.users['email'],
  creation_time: osm.users['creation_time']
}>`SELECT (email, creation_time) FROM users WHERE creation_time >= '2013-01-01'`;
```
(from https://www.cs.mcgill.ca/~mxia3/2016/11/18/Statically-typed-PostgreSQL-queries-and-typescript-schemats/#Writing-code-with-typed-schema)

karlhorky:
@Minigugus @porsager what is left over before this can be merged?

Would love to start using a typed version of postgres.js, even if the type info is missing some things that would be nice to have.

Can we merge a first version and then come back to some of the "nice to have"s later?

Like @cfanoulis I would also like to help at that stage, contributing some parts for the "nice to have" bits.

Minigugus:
@karlhorky 
> Would love to start using a typed version of postgres.js, even if the type info is missing some things that would be nice to have.

Currently, I think types are in sync with the JavaScript API. In fact, I'm testing them with the `tests/index.js` file, which covers the whole public API. I don't encourage you but if you really need them now, you can add them manually in your project, like [some people do](https://github.com/yckao/verdaccio-pg-storage/blob/master/types/postgres.d.ts). :+1: 

> what is left over before this can be merged? [...] I would also like to help at that stage, contributing some parts for the "nice to have" bits.

I think the only missing part is the JSDoc. I'm not a native English speaker and I don't have time to work on it, so if you want to help, you are welcome :sweat_smile:

> Eg. Can we merge this without full JSDoc types? [...] Would love to start using a typed version of postgres.js

@porsager What about a `postgres@beta` version? Not only for types but also for working v2 features?

karlhorky:
> Currently, I think types are in sync with the JavaScript API
> I don't have time to work on it

Ok, then since you don't have time to work on it, seems like it doesn't make sense to wait for it, since the types are already there?

@porsager Maybe the current version can be merged and released with v2? How far away is v2?

Then after the merge a new pull request opened for the JSDoc / TSDoc descriptions.

porsager:
@Minigugus surely, I should have thought of that long ago. Let's look at merging this later today and then releasing a @beta version.

@karlhorky There is still a little bit left to do for v2, but I certainly hope that it's ready within the next couple of weeks.

porsager:
Okay @Minigugus, it all looks great to me from my limited typescript experience, are you fine with merging now? :)

Minigugus:
@porsager Of course I am :+1: :grin:

porsager:
Awesome.. Thanks a lot @Minigugus !!

karlhorky:
Nice one, thanks @Minigugus and @porsager!

karlhorky:
@Minigugus so I want to open a new pull request for contributing the JSDoc types. Just so that we are on the same page, I assume that you would like this level of documentation:






...for things like this:





Is that what you meant in https://github.com/porsager/postgres/pull/8#issuecomment-572480796 ?

porsager:
@Minigugus There's now a postgres@beta npm release ;)

https://github.com/porsager/postgres/releases/tag/v2.0.0-beta.0

Minigugus:
@karlhorky Exactly :+1: The idea is to get something like [that](https://github.com/sequelize/sequelize/blob/8244a34eaa41d40c2170eec6c9b6cf5ff4378e38/types/lib/sequelize.d.ts#L76-L115), with [examples if possible](https://github.com/sequelize/sequelize/blob/8244a34eaa41d40c2170eec6c9b6cf5ff4378e38/types/lib/sequelize.d.ts#L801-L832) for complex parts like `const sql = postgres()` for instance :wink:

@porsager Nice :+1: Now we just have to wait for the feedback from the users :laughing:

xiaoyu-tamu:
@Minigugus Looks like the return type of sql``; is not correct? Or I did something wrong?

image

image

yckao:
> Should it be `{id:number}[]` instead of `{id:number}`

sql`` will always return a array, so in the generics type will only use as an object.
also [user1] = sql`...` is an destructuring assignment, so it will be `{ id: number }`

If you want get all users, instead of first user, you should simply remove this [].
like `const user = await sql< id: number>`SELECT * FROM user``

xiaoyu-tamu:
Ah, I was about to delete my previous comment, I just figure out that. Thanks

Minigugus:
@xiaoyu-tamu @yckao Too late :laughing: I still send my comment in case someone else as this question :wink:  

It is not clear of the type here represent the column type or the resulting array, so I managed to support both as PostgreSQL does not permit arrays directly on top level (there will always be column names at top level). However, only the first top level array is removed, so you can still pass `{ id: number }[][]` even if it will never be the correct type (except for `sql.unsafe` which permit multi-statements and then returns something like `[[q1r1, q1r2], [q2r1, q2r2], ...]`).

Another advantage of keeping the array is that is permit to pass tuples for queries with statically known row number, like with ```const [user, incorrect] = await sql<[{ id: number }]>`SELECT * FROM users LIMIT 1`;``` or to pass different types per columns for instance :+1: *As of now, there is a TypeScript bug that prevent the previous exemple to report that `incorrect` is outside of the tuple, so you have to pass `[{ id: number}, never]` instead.*

xiaoyu-tamu:
@Minigugus 

How would you handle undefined? The query may return undefined if no rows match. Currently, sql<{} | undefined> is now allowed.

e.g.

```const [firstUser] = sql<{ id: number }>`SELELCT  * FROM user WHERE condition LIMIT 1`;```

**Comment by yckao** [comment]
@Minigugus
It seems like we can support undefined when using tuple.
Like
https://github.com/yckao/postgres/commit/0beec295f49f58ae4badd4e4c3f5b6fdf488bc74

Use like 
`const [user] = sql<[{ id: number } | undefined]>`SELELCT * FROM user WHERE condition LIMIT 1`;`

@xiaoyu-tamu
Maybe you can open an issue, and we can discuss further on it.
Let's leave this thread clean.

**Comment by karlhorky** [comment]
> The query may return undefined if no rows match

@xiaoyu-tamu @yckao Wouldn't the type already work without any special `undefined` annotation?

Eg. the following code typechecks just fine:

```typescript
type User = {id: number}
const arr: User[] = []
```

Try it on the playground: https://www.typescriptlang.org/play/?ssl=2&ssc=23&pln=1&pc=1#code/C4TwDgpgBAqgzhATlAvFA3gSwCYC4oB2ArgLYBGSAvgFADGA9gXMFAIaKL7xIDaAuqij8gA

yckao:
@karlhorky Yes, the type is already work and doesn't do anything wrong.

But the issue @xiaoyu-tamu mention could be a interest enhancement.

The issue is the zero length array in the typescript with destructuring assignment will not indicate it as undefined.
And the usage is a very common usage when only fetch first row.
Also this usage is defined as tuple in the definition
So something like this is possible.
```const [user] = sql<[{ id: number } | undefined]>`SELELCT * FROM user WHERE condition LIMIT 1`;```

We can let user decide whether the first row could undefined or not.

So I think this is not a bug, but can be an enhancement.

Also I think we should create an issue instead discussing in this thread since it is already merged.

PR #9, opened 2020-01-10
Improved tag function

infusion:
Hi Rasmus,

I read over the library and this is really an amazing work! Since you use templates, you could also use the new spread operator to speed things up a bit.

Robert

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -55,8 +55,8 @@ function Postgres(url, options) {
     , slots = max
     , listener
 
-  function postgres(xs) {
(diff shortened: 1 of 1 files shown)


porsager:
HI Robert.

Thanks a lot :)

I actually removed all rest/spread usage before 1.0 because I was adding support for Node 6.x, but it seems my assumption was wrong and that it is 5.x which doesn't support it.

I'll just have to check if there was another reason, but if not this should be good ;)

porsager:
Just for kicks I tried to do some performance tests on these 2 variations, and there is really no difference.

I'll hold off on adding any rest / spread usage for now.

PR #34, opened 2020-02-14
Fix hanging issue with sql.end

### LINKED ISSUE #32: sql.end({timeout: null}) hangs
Hi and thanks for making this driver available. I'm evaluating it as a replacement to pg-promise.

Our tests hangs in the teardown phase if we end the connection with `sql.end({timeout: null})`. This is due to the implementation of `socket.end` available in version 1.0.2
```
    end: () => {
      return new Promise(r => socket && socket.end(r))
    }
```
If there's no socket, `r` is never called.

This has been fixed in 4156c3f2fddf5c6694a3877b5d3516f16f55ba48 but it hasn't been released yet. Can you please release a new version?

ffissore:
When closing a connection, check if either `ready` is true or there's no running query
Fixes #32

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -91,7 +91,7 @@ function Connection(options = {}) {
       ended = () => resolve(socket.end())
     })
 
-    process.nextTick(() => ready && ended())
+    process.nextTick(() => (ready || !backend.query) && ended())
 
     return promise
   }


porsager:
Hi @ffissore, so sorry for the long wait here!

I'd like to have a test that can show the issue, and that this issue fixes it. Would you mind adding that?

porsager:
The connection handling has changed a bit in the latest on master, so this one should be solved there. Let me know if that's not the case on your end. Thaks.

PR #58, opened 2020-04-17
Add quotes to escaped parameters

### LINKED ISSUE #57: Quote all columns in generated inserts, updates
I use camelCase for table and column names.  These need to be quoted.  Currently when I do something like:

```
    await sql`
      INSERT INTO "Address" ${sql(address)}
    `
```

The fields aren't being quoted.

naturalethic:
Fixes #57

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -449,7 +449,7 @@ function Postgres(a, b) {
   }
 
   function escapeHelper(xs) {
-    return xs.reduce((acc, x) => acc + (acc ? ',' : '') + escape(x), '')
+    return xs.reduce((acc, x) => acc + (acc ? ',' : '') + '"' + escape(x) + '"', '')
   }
 
   function parseValue(x, xargs, types) {


porsager:
This breaks other things, and there are also failing tests. I realize I didn't have a test case for the casing, but it should already be handled, and quoted if needed (if uppercase characters are included the column name will be quoted.)

Let's debug further in the issue if you're having problems ;)

porsager:
So the fix needs to be done in the escape serializer and there add quotes if there are any uppercase characters. I actually thought that's how I had done it, but I see I'm missing a test case for it, so gotta look closer if it's a regression or if I simply remember wrong.

porsager:
A fix would be to add ` || str !== str.toLowerCase()` at the [end of line 70 in the escape function](https://github.com/porsager/postgres/blob/master/lib/types.js#L70)

Maybe add a test like 
```
t('dynamic insert quoting', async() => {
  await sql`create table test ("lowAndUpper" int)`
  const x = { lowAndUpper: 42 }

  return [42, (await sql`insert into test ${ sql(x) } returning *`)[0].lowAndUpper, await sql`drop table test`]
})
```

naturalethic:
Can you give me a copy of your postgres configs (redacted where necessary) to aid me in setting up the proper ssl environment for your tests, plus any other relevant info I might need?

porsager:
Oh I'm sorry, I completely forgot there were environment specific things missing to run the test 🤦‍♂️

I'll get it added to the README.md while I'm at it.

I'd want to set up github actions to run at some point too..

Issue #78, opened 2020-05-28
Please change license from WTFPL to something supported by FSF/OSI


galniv:
Looking at the mess that's software licenses, I 100% get the "fuck this" approach to dealing with them :) that being said, this license has some issues, specifically with warranty, the vague language of the terms grant, and the fact that "public domain" licenses don't really legally exist in various countries. All of these have led the [FSF](https://www.gnu.org/licenses/license-list.html#WTFPL) and the [OSI](https://opensource.org/minutes20090304) to reject it as a good open source license, and also led to companies not being able to use it for silly, but unfortunately real legal reasons (I don't work there, but [Google](https://opensource.google/docs/thirdparty/licenses/#wtfpl-not-allowed) is one clearly documented example).

Would you consider moving the library to a more established license that's just as free? It could even be the [Unlicense](https://unlicense.org/), which is public domain but with a "more legal" fallback way of saying "do what you want".

porsager:
Hi @galniv.

Yeah, it's not easy to be against intellectual property these days, but I also haven't updated myself in a long while. I think your suggestion of going with Unlicense seems good. I'll do a bit more reading too, but I'm fine changing it to something better!

galniv:
Great, thanks a lot! I'm also happy to create a PR for Unlicense or whatever else you choose, if that makes things easier.

porsager:
Hi @galniv ..  Sorry, got caught up in work. You're very welcome to submit a PR for the change.

galniv:
Thanks a lot for your willingness to adapt the license, and for the quick turnaround in merging the PR! Closing this 🎉

PR #80, opened 2020-06-11
Update license to the Unlicense

galniv:
As discussed in https://github.com/porsager/postgres/issues/78, this PR changes the repo's license to the [Unlicense](https://unlicense.org/), which is a public domain license with a "more legal" fallback way of saying "do what you want" than the one offered by WTFPL.


porsager:
Thanks a lot @galniv !

PR #82, opened 2020-06-20
add support for named parameters

ksteckert:
To further reduce mistakes in "unsafe" queries, I've added support for named variables.
The `paramify` function parses the query and replaces named params with indexed params.

```js
let dynamically_built_sql = 'select * from users where id = $id'
let bindings = {id: 5}
await sql.unsafe(dynamically_built_sql, bindings)
// converts & runs query as: 'select * from users where id = $1'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,6 +543,20 @@ If you know what you're doing, you can use `unsafe` to pass any string you'd lik
 sql.unsafe('select ' + danger + ' from users where id = ' + dragons)
 
 ```
+
+Indexed parameters will be escaped by postgres.
+```js
+
+sql.unsafe('select ' + danger + ' from users where id = $1', [dragons])
+
+```
+
+Named parameters may contain letters, numbers, and underscores, however they must start with a letter.  Named parameters get converted to indexed parameters, which will be escaped by postgres.
+```js
+// $var_1 is valid, but $1_var is invalid
+sql.unsafe('select ' + danger + ' from users where id = $id', {id: dragons})
+```
+
 </details>
 
 ## Errors
--- a/README.md
+++ b/README.md
@@ -540,21 +540,32 @@ If you know what you're doing, you can use `unsafe` to pass any string you'd lik
 
 ```js
 
-sql.unsafe('select ' + danger + ' from users where id = ' + dragons)
+await sql.unsafe('select ' + danger + ' from users where id = ' + dragons)
 
 ```
 
 Indexed parameters will be escaped by postgres.
 ```js
 
-sql.unsafe('select ' + danger + ' from users where id = $1', [dragons])
+await sql.unsafe('select ' + danger + ' from users where id = $1', [dragons])
 
 ```
 
-Named parameters may contain letters, numbers, and underscores, however they must start with a letter.  Named parameters get converted to indexed parameters, which will be escaped by postgres.
+Named parameters may contain letters, numbers, and underscores, however they must start with a letter.  Named parameters get converted to indexed parameters, which will be escaped by postgres.  If the parameters object does not contain all expected keys, null values will be used.
 ```js
 // $var_1 is valid, but $1_var is invalid
-sql.unsafe('select ' + danger + ' from users where id = $id', {id: dragons})
+await sql.unsafe('select ' + danger + ' from users where id = $id', {id: dragons})
+```
+
+You may also call `sql.paramify(query, [args], strict = false)` directly if you want to enforce strict parameter naming.  With `strict` enabled, `postgres` will throw an error if the object does not contain all the expected keys.  This can help prevent naming errors.
+```js
+let params = sql.paramify('select ' + danger + ' from users where id = $id', {no: dragons}, true)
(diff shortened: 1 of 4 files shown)


ksteckert:
This PR has the same exact content as PR #83, except this is for version 2.

ksteckert:
I've noticed that in `postgres` you allow the use of `sql(obj, 'field1', 'field2')` even if there are no keys for 'field1' or 'field2', by using null values.  This is handy for inserts when those fields may not be present in an object.  If you want to maintain that same behavior with this update, just update paramify.js with the following simplfied function instead.

```js
function Paramify(sql, parameters = {}) {
  let sqlParams = sql.match(paramPattern).map(t => t.substring(1)).filter((val, ix, arr) => arr.indexOf(val) === ix)
  let text = sqlParams.reduce((reducedSqlQuery, param, ix) => {
    let paramPattern = new RegExp('\\$' + param + '\\b', 'g')
    return reducedSqlQuery.replace(paramPattern,'$' + (ix+1)) // pg params are 1-indexed
  }, sql)
  let values = sqlParams.map(p => parameters[p])

  let out = {text, values}
  return out;
}
```

ksteckert:
I've noticed that in `postgres` you allow the use of `sql(obj, 'field1', 'field2')` even if there are no keys for 'field1' or 'field2', by using null values. This is handy for inserts when those fields may not be present in an object.  

I set `sql.unsafe()` to use this behavior by default for named parameters, while exposing a `sql.paramify()` function for those that want to ensure the parameter object contains all the necessary keys.

I documented the usage examples in `readme.md`.

porsager:
Hi @ksteckert .

Thanks for your interest in Postgres.js. 

I'm not sure I'd want to encourage more use of sql.unsafe by adding named parameters like this though.

Could you maybe try to describe why ``` sql`` ``` isn't sufficient?

ksteckert:
There are many times when I build functions that create SQL code., Sometimes, I may want to add a condition based on certain criteria, and it's more efficient to do it in code than using a CASE in SQL.

Obviously, this example below is *very* simple and could be resolved with an OR.  Imagine a much larger, but similar in nature, real-world query that's hundreds of lines long, maybe for an reporting platform, where you're constructing the report SQL in code, but you want to use named parameters for security purposes.  Lots of conditional concatenating.
```
let something_else = (criteria === some_val) ? 'username = $username' : 'email= $email'
let query = `
  select * from users
  where not deleted
    ${something else}
`
```

ksteckert:
Also there are times when I have a LARGE block of SQL that is used for a bunch of different queries.
Other times I need to prepare a SQL statement, and then execute it with a binding object any number of times.  I realize I could create a function for this... but putting each query in it's own function would be tedious.  I'd rather just be able to call sql.unsafe(prepared_sql, binds) when I have that use case scenario.

I really like this, and I'd like to be able to migrate to it for my node projects.  Even if you don't implement this feature, I could still use my parmify function locally and call unsafe(), but it'd be nice if I could pass an dictionary object just as easily as I could pass an array of values.

ksteckert:
I'd love to show you some real-world SQL examples, but all the ones I have are covered by non-disclosure agreements.  :-D

Using named parameters isn't unsafe, but there's the possibility that someone doesn't use them correctly; I understand that.  Tagged templates force you to use it safely, you can't accidentally put a variable in there that doesn't get escaped.  I've started using tagged templates all over the place for some of my smaller projects.  My larger projects would require the use of unsafe() in a handful of scenarios, and I'd prefer to have that functionality baked in.  Adding $1 $2 $3 to SQL isn't very intuitive, names are much easier to read.  I built a solution for myself, and just figured I'd pass it along to the community.

ksteckert:
All that said, I'm very impressed by `postgres.js`!  It blew me away how simple it is to use, and that it has 0 dependencies!!!  I can't wait to use it in even more projects.

The `unsafe()` section is buried so deep in there I don't think anyone would feel 'encouraged' use it over tagged templates.  Most people will probably never even see that section unless they are specifically looking for it like I did.  I only found it because I was digging through the source code trying to figure out how to run a SQL statement stored in a variable.

Minigugus:
@porsager is right: you should adapt ```sql``;``` instead of `sql.unsafe`.

You can for instance add a ```sql.escape``;``` function that returns an helper like `sql(obj, ...)` :

```ts
let something_else = (criteria === some_val) ? sql.escape`username = ${username}` : sql.escape`email= ${email}`
let query = sql`
  select * from users
  where not deleted
    and ${something_else}
`
```
Currently, `sql.array`, `sql.json`, `sql('str', ...)` and so on does not return `string` but [objets](https://github.com/porsager/postgres/blob/806cc2c3a42e453a7fb5292989a74f77a1eeda72/types/index.d.ts#L298-L304) in order to be properly escaped in ```sql``;```.

porsager:
@ksteckert Sorry for the missing followup on this, and thank you for the kind words.

I think this is better as a standalone module considering how simple it would be to use with Postgres.js. That way we can keep the core simpler.

PR #83, opened 2020-06-20
add support for named parameters (postgres v1)

ksteckert:
To further reduce mistakes in "unsafe" queries, I've added support for named variables.
The paramify function parses the query and replaces named params with indexed params.

let dynamically_built_sql = 'select * from users where id = $id'
let bindings = {id: 5}
await sql.unsafe(dynamically_built_sql, bindings)
// converts & runs query as: 'select * from users where id = $1'

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,15 @@ let params = sql.paramify('select ' + danger + ' from users where id = $id', {no
 let params = sql.paramify('select ' + danger + ' from users where id = $id', {id: dragons}, true)
 await sql.unsafe(...params)
 ```
+
+```js
+let params = sql.paramify('update users set first_name = $fname where id = $id', {id: 5}, false)
+await sql.unsafe(...params)
+// This will updated the first_name to `null`
+// This is the default behavior for sql.unsafe when using named parameters.
+await sql.unsafe('update users set first_name = $fname where id = $id', {id: 5})
+```
+
 </details>
 
 ## Errors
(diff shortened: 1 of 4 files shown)


ksteckert:
This PR has the same exact content as PR #82, except this is for version 1.

ksteckert:
I've noticed that in `postgres` you allow the use of `sql(obj, 'field1', 'field2')` even if there are no keys for 'field1' or 'field2', by using null values.  This is handy for inserts when those fields may not be present in an object.  If you want to maintain that same behavior with this update, just update paramify.js with the following simplfied function instead.

```js
function Paramify(sql, parameters = {}) {
  let sqlParams = sql.match(paramPattern).map(t => t.substring(1)).filter((val, ix, arr) => arr.indexOf(val) === ix)
  let text = sqlParams.reduce((reducedSqlQuery, param, ix) => {
    let paramPattern = new RegExp('\\$' + param + '\\b', 'g')
    return reducedSqlQuery.replace(paramPattern,'$' + (ix+1)) // pg params are 1-indexed
  }, sql)
  let values = sqlParams.map(p => parameters[p])

  let out = {text, values}
  return out;
}
```

ksteckert:
I've noticed that in `postgres` you allow the use of `sql(obj, 'field1', 'field2')` even if there are no keys for 'field1' or 'field2', by using null values. This is handy for inserts when those fields may not be present in an object.  

I set `sql.unsafe()` to use this behavior by default for named parameters, while exposing a `sql.paramify()` function for those that want to ensure the parameter object contains all the necessary keys.

I documented the usage examples in `readme.md`.

PR #84, opened 2020-06-20
Better sql`` types

Minigugus:
1. Set row type to `YourType | undefined` in ```sql<YourType>``;``` but to `YourType` in ```sql<YourType[]>``;``` so that the user can choose the best regarding the situation (see also https://github.com/porsager/postgres/pull/8#issuecomment-646630154) :
```ts
const [user] = await sql<User>`SELECT * FROM users WHERE id=${id}`;
user.name // error : `user` can be `undefined`
user?.name // ok

for (const user of await sql<User[]>`SELECT * FROM users WHERE id=${id}`)
  user.name // ok since `user` will always be defined

await sql<User>`SELECT * FROM users WHERE id=${id}`.cursor(user => {
  user.name // ok (same as `for`), supported by both the `User` and the `User[]` form
});

const [user] = await sql<[boolean]>`SELECT * FROM users WHERE id=${id}`;
user.valueOf() // incorrect but ok (the tuple form accept any type)
```
2. Currently, primitive types are not permitted as row types, even in ```sql<number>`SELECT 1`;```. This PR add `T` to `{ '?column?': T }` conversion if `T` is `Serializable` (`string`, `number`, `Date`, ...) :
```ts
const [{ '?column?': value }] = await sql<number>`SELECT ${nb}`; // ok
// => value: number
```

*Feel free to discuss types here before the release :+1:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,46 @@ const [new_user] = await sql`
 // new_user = { user_id: 1, name: 'Murray', age: 68 }
 ```
 
+
+#### TypeScript support
+
+`postgres` has TypeScript support. You can pass a row list type for your queries in this way:
+```ts
+interface User {
+  id: number
+  name: string
+}
+
+const users = await sql<User[]>`SELECT * FROM users`
+users[0].id // ok => number
+users[1].name // ok => string
+users[0].invalid // fails: `invalid` does not exists on `User`
+```
+
+However, be sure to check the array length to avoid accessing properties of `undefined` rows:
+```ts
+const users = await sql<User[]>`SELECT * FROM users WHERE id = ${id}`
+if (!users.length)
+  throw new Error('Not found')
+return users[0]
+```
+
+You can also prefer destructuring when you only care about a fixed number of rows.
+In this case, we recommand you to prefer using tuples to handle `undefined` properly:
+```ts
+const [user]: [User?] = await sql`SELECT * FROM users WHERE id = ${id}`
+if (!user) // => User | undefined
+  throw new Error('Not found')
+return user // => User
+
+// NOTE:
+const [first, second]: [User?] = await sql`SELECT * FROM users WHERE id = ${id}` // fails: `second` does not exist on `[User?]`
+// vs
+const [first, second] = await sql<[User?]>`SELECT * FROM users WHERE id = ${id}` // ok but should fail
+```
+
+All the public API is typed. Also, TypeScript support is still in beta. Feel free to open an issue if you have trouble with types.
+
 #### Query parameters
 
 Parameters are automatically inferred and handled by Postgres so that SQL injection isn't possible. No special handling is necessary, simply use JS tagged template literals as usual.
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -244,6 +244,24 @@ declare namespace postgres {
     [column: string]: any;
   }
 
+  interface UnlabeledRow<T = any> {
+    '?column?': T;
+  }
+
+  type MaybeRow = Row | undefined;
+
+  type TransformRow<T> = T extends Serializable
+    ? { '?column?': T; }
+    : T extends MaybeRow
+    ? T
+    : Row;
+
+  type AsRowList<T> = T extends any[]
+    ? readonly any[] extends T
+    ? TransformRow<T[number]>[]
+    : T
+    : TransformRow<T | undefined>[];
+
   interface Column<T extends string> {
     name: T;
     type: number;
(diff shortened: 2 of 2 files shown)


yckao:
Looks good.
But maybe add some example and docs into README will be better since the difference between sql<User> and sql<User[]> maybe not very intuitive.

yckao:
In my opinion, maybe only set row type to `YourType | undefined` when using tuple `sql<[YourType]>` will be more intuitive.
And if we can use `sql<[YourType]>` and `sql<YourType[]>`, maybe remove `sql<YourType>` will be less confusion.

Minigugus:
> (Just because we could write something like `` const [user] = await sql<[User]>`SELECT * FROM users WHERE id=${id}`; ``, and `[user]` is like `[User]` :laughing:)

I agree but currently it's not possible in TypeScript to transform types in tuples without loosing types positions (i.e. `[A, B]` => `[A | undefined, B | undefined]` not possible, only => `(A | B | undefined)[]`). Also, in my opinion, when we pass `[User]` I think we expect to get `[User]` as a result, not `[User | undefined]`. I wanted to keep tuples for complex use cases not compatible with other forms.

But in the meantime I found [another possible solution](https://www.typescriptlang.org/play/#code/C4TwDgpgBAggPAFQHxQLxQBQKgHygVwDsATCAMwEtCJiBKAbQF0oAyKASWAgCcBDAIwA2ETj17AA9t0RIA3ACh5pAMaDe3aMomEAzsCi8AXLDh7uVAOZzFW3fvr9m6XgvlkpmW3qjKoEsga0UADeigC+irz0AAyMAHS8hMoAFlJAA) :laughing: (hover `a` and `b`)

yckao:
> I agree but currently it's not possible in TypeScript to transform types in tuples without loosing types positions (i.e. `[A, B]` => `[A | undefined, B | undefined]` not possible, only => `(A | B | undefined)[]`).

Partial will work with tuple.
[Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=5&pc=11#code/C4TwDgpgBAggTnKBeKBtAzsOBLAdgcwBooBvXAQwFsIAuKTHAgX2NwFdKAjCOVtgG34BdAFChIUACoRMyKAAVycYNnL8APPDgA+ESIDGAe1yzydabJSoA5OmvESUCtTrXynfdagsobXABMIADM8CH8+QVFyVABGIQB+ADpnCCA)

And just wondering if ```const [user] = await sql<[User | undefined]>`SELECT * FROM users WHERE id=${id}`;``` will work with new typing?

Minigugus:
> Partial will work with tuple

Oh nice that's what I was looking for :+1: 

> And I think tuple usage is more fit in this situation, because most situation we need to handle undefined like this is when we using destructuring with known length of elements, and is same to the usage of tuple.
Is there any other complex usage you mention above I missed?

What about ```const [{ n }] = sql<[{ n: 42 }]>`SELECT 42`;```? Here `n` will *always* be `42`, not  `42 | undefined`. My opinion is that we should let the user decide where a row is defined or not.

> Also just wondering ```if const [user] = await sql<[User | undefined]>`SELECT * FROM users WHERE id=${id}`;``` will work with new typing?

It should as tuples are passed as is. :wink:

yckao:
> What about `` const [{ n }] = sql<[{ n: 42 }]>`SELECT 42`; ``? Here `n` will _always_ be `42`, not `42 | undefined`. My opinion is that we should let the user decide where a row is defined or not.

That seems reasonable.
And next problem is if the usage of `sql<User>` is reasonable.

I seems it may be a little bit confusion.
Since we can simply achieve the usage with
```const [user] = await sql<[User | undefined]>`SELECT * FROM users WHERE id=${id}`;```
or
```const [user] = await sql<Partial<[User]>>`SELECT * FROM users WHERE id=${id}`;```

And I also notice that the type of 
```const [user] = await sql<[User | undefined]>`SELECT * FROM users WHERE id=${id}`;```
is currently `User` not `User | undefined` even in this commit.

**Comment by Minigugus** [comment]
@yckao So you suggest to drop support for `sql<User>` with a contraint like `T extends any[]`? I think not being able to pass `sql<User>` *is* confusing (and I cannot reject arrays as they are infered as objets), but maybe I'm wrong.

What about a survey :laughing:? React to this message with:
 * :+1: for `sql<User[]>` only
 * :-1: for `sql<User>` only
 * :confused: for both
 * (none) for anything else

**Comment by Minigugus** [comment]
Ok finally I think you're right: `sql<User[]>` is better for users but also for TypeScript (infer better):
```ts
const [maybeUser]: [User?] = await sql`SELECT * FROM users WHERE id = ${id}`; // ok
// => maybeUser: User | undefined

// @ts-expect-error
const [invalid] = await sql<42>`SELECT 42`; // fails: `42 not assignable to any[]`
const [valid]: [undefined] = await sql`SELECT null`; // ok but useless
// => valid: undefined
```
But `undefined` is still not enforced in case the query cannot returns 0 rows, or if the user check array bounds correctly:
```ts
// Tuple mapping now works with the `any[]` constraint on `T`:
// `[42]` mapped to `[{ '?column?': 42 }]`
const [{ "?column?": n }] = await sql<[42]>`SELECT 42`; // ok
// => n: 42

const result = await sql<User[]>`SELECT * FROM users WHERE id = ${id}`; // ok
if (!result.length)
  throw new Error('Not found');
return result[0]; // => result[0]: User

const result = await sql<User[]>`SELECT * FROM users WHERE name LIKE ${pattern}`; // ok
return result.map(user => user.id);
```
<hr>
To conclude, I think the following form is peferable when destructuring:

```ts
const [user, maybeUser]: [User, User?] = await sql`...`; // ok
// => user: User
// => maybeUser: User | undefined

const [first, second, third]: [User, User?, User] = await sql`...`; // fails
//                                        ^^^^^^ 
//           A required element cannot follow an optional element

const [first, second, third]: [User, (User | undefined)[]] = await sql`...`; // ok
// => first: User
// => second: User | undefined
// => third: User | undefined
```
For any other case, ``sql<YourType[]>`...`;`` works fine.

I'll push changes soon, stay tuned :wink:

PR #90, opened 2020-06-25
import/export syntax in readme

nainardev:
Node.js now allows import/export syntax.
Also you use top-level await, so i assume this change is welcome

Also in readme, import/export syntax is used in [asas](https://github.com/porsager/postgres/blob/master/README.md#sample-shutdown-using-prexit)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,23 +14,21 @@
 <br>
 
 **Install**
-```bash
+```sh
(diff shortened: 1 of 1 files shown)


porsager:
Thanks @bnainar 

I think we'll wait to clean this up when making proper es module support.

PR #99, opened 2020-08-01
Expose the PostgresError type for TypeScript

akheron:
With current `index.d.ts` it seems to be impossible to get hold of the `PostgresError` type. It would be useful for e.g. signatures of error handling functions.

Fix by exposing the type but not the class.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -93,7 +93,10 @@ type UnwrapPromiseArray<T> = T extends any[] ? {
   [k in keyof T]: T[k] extends Promise<infer R> ? R : T[k]
 } : T;
 
+type PostgresErrorType = PostgresError
+
 declare namespace postgres {
+  type PostgresError = PostgresErrorType
 
   /**
    * Convert a string to Pascal case.


types/index.d.ts
```diff
@@ -63,7 +63,7 @@ interface JSToPostgresTypeMap {
   [name: string]: unknown;
 }
 
-declare class PostgresError extends Error {
+declare class PostgresErrorImpl extends Error {
```
Minigugus, line 66:
```suggestion
declare class PostgresError extends Error {
```

types/index.d.ts
```diff
@@ -94,6 +94,7 @@ type UnwrapPromiseArray<T> = T extends any[] ? {
 } : T;
 
 declare namespace postgres {
+  type PostgresError = PostgresErrorImpl
```
Minigugus, line 97:
```suggestion
  type PostgresError = typeof PostgresError
```

types/index.d.ts
```diff
@@ -312,7 +313,7 @@ declare namespace postgres {
     <T extends HelperSerializable, U extends (keyof (T extends any[] ? T[number] : T))[]>(objOrArray: T, ...keys: U): Helper<T, U>;
 
     END: {}; // FIXME unique symbol ?
-    PostgresError: typeof PostgresError;
+    PostgresError: typeof PostgresErrorImpl;
```
Minigugus, line 316:
```suggestion
    PostgresError: PostgresError;
```

akheron:
I don't think this is right. I need the type of `PostgresError` instances. `typeof PostgresError` gives the type of the constructor function.

akheron:
I think the easiest way would be to move `class PostgresError { ... }` inside the namespace, but I guess that was not done on purpose?

Minigugus:
> I don't think this is right. I need the type of PostgresError instances. typeof PostgresError gives the type of the constructor function.

Indeed, sorry :laughing:. Maybe a `type PostgresErrorType = PostgresError;` before the `declare namespace`, and `type PostgresError = PostgresErrorType` after? :neutral_face: 

> I think the easiest way would be to move class PostgresError { ... } inside the namespace, but I guess it was left outside on purpose?

Yes, because in this case it becomes possible to access the class constructor from the package exports, which is wrong: it is only accessible on the `sql` tag.

akheron:
> Maybe a type PostgresErrorType = PostgresError; before the declare namespace, and type PostgresError = PostgresErrorType after? :neutral_face:

Fixed like this ^

porsager:
I guess this is good to merge - right @Minigugus ?

Minigugus (approved):

Issue #12, opened 2020-01-10
Better dynamic query building


okbrown:
Hi, First and foremost, I think this is a great start at making something really simple and effective. Thank you!!

However... I have always found template strings a little odd especially in this context.
Let us say we have

``const query = `SELECT ${distinct}, ${columns} FROM ${table} ${innerJoins} WHERE ${where}`;``
(Where distinct, columns, table etc are their own strings made up of other functions elsehwhere)
which basically outputs...

```
SELECT DISTINCT(person.uid), person.name, movie.title, movie.year
FROM person 
INNER JOIN movie_ref ON movie_ref.person_id = person.id
INNER JOIN movie ON movie.id = movie_ref.movie_id
WHERE person.name ILIKE '%luke%'
```

I should then be able to do ``sql`${query}`.then((data) => console.log(data)`` or `await` etc..

But instead I receive a syntax error:
`Error: syntax error at or near "$1" at Object.generic (../node_modules/postgres/lib/types.js:200:5)`

Gonna take a look at how this tagged template function is actually working as Im aware it takes the values $1, $2 etc.. and so forth to evaluate them and pass them to the statement e.g. `select * from table where id = $1` => `select * from table where id = somevalue`

At the moment, it means I have to directly place all my select statements big or small in this method. Which will start looking very messy especially when queries become more complex and the app grows.

porsager:
Hi.

Thanks :)

Yeah, it can take a while to wrap your head around tagged template functions, but once you do you start seeing great usage for them everywhere (at least i did) ;).

The first thing is to understand the difference between [tagged template functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates) and just regular template literal strings. 

Tagged template functions allows you to handle the values `${ value }` and ensure they are escaped or translated properly. If you were able to concat strings together without any handling you would quickly open up for various kinds of sql injection, which is also why you have to do ``` sql`...` ```, and are not able to do ``` sql(`...`) ```, when making queries.

Postgres.js converts normal values and replaces their position in the query with `$1, $2, etc` and send the actual values as parameters to Postgres over the wire, so that Postgres can do its magic from there on. (you can add `debug: console.log` to the options to see how the queries end up.)

That mens you can't pass keywords and identifiers directly as raw string arguments when calling the `sql` function as a tagged template function (``` sql`` ```) since that would be just as dangerous as simply contatting strings..

[...]

PR #101, opened 2020-08-01
Add sql.partial``

yckao:
related to issue #12 

Maybe there will be other better way.
Just create a PR and want to have a discussion.

Can't wait to have feature like this since partial is essential in my use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,15 @@ const [{ json }] = await sql`
 // json = { hello: 'postgres' }
 ```
 
+#### Partial ``` sql.partial`` ``` and  Skip ``` sql.skip() ```
+With  ``` sql.partial`` ``` you can pass partial query into ``` sql`` ``` that can use for dynamic query building. And use ``` sql.skip() ``` if there is no partial query needed.
+
+```js
+
+await sql`select * as user from users ${id ? sql.partial`where id = ${id}` : sql.skip() }`
+
+```
+
 ## File query `sql.file(path, [args], [options]) -> Promise`
 
 Using an `.sql` file for a query. The contents will be cached in memory so that the file is only read once.
--- a/lib/index.js
+++ b/lib/index.js
@@ -235,6 +235,19 @@ function Postgres(a, b) {
     }
   }
 
+  function partial(xs) {
+    const o = Object.create(notPromise)
+    o.partial = xs
+    o.args = Array.from(arguments).slice(1)
+    return o
+  }
+
+  function skip() {
+    const o = Object.create(notPromise)
+    o.skip = true
+    return o
+  }
+
   function fetchArrayTypes(connection) {
     return arrayTypesPromise || (arrayTypesPromise =
       new Promise((resolve, reject) => {
@@ -274,7 +287,9 @@ function Postgres(a, b) {
       unsafe,
       array,
       file,
-      json
+      json,
+      partial,
+      skip
     })
 
     function notify(channel, payload) {
@@ -391,10 +406,7 @@ function Postgres(a, b) {
     }
   }
 
-  function parse(query, xs, args = []) {
-    const xargs = []
-        , types = []
-
+  function parse(query, xs, args = [], xargs = [], types = []) {
     let str = xs[0]
     let arg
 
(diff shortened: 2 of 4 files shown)


Minigugus:
Can just ```sql``;``` and ```sql.partial``;``` be the same?

I mean I know that ```sql``;``` returns a promise and therefore executes the query, but it might return an object with a `.then` function that actually executes the query, like a fake promise (works with await). Therefore the partials would simply be```sql``;``` not awaited:
```js
await sql`SELECT id FROM products ${price ? sql`WHERE price >= ${price}` : sql``}`;
```
Moreover, queries would also become reusable :wink:

yckao:
@Minigugus Thanks.

I would like to keep using sql.partial because sql`` is already have so many duty, and I don't want increase complexity.
So I implemented sql.partial in a simplest way.

If let ``` sql`` ``` equal to ``` sql.partial`` ``` it will also effect the execute order with `send` and others because of the timing we parse value. That will be more complicated.

So I will keep sql.partial in this PR to keep it simple.

porsager:
Hi @yckao 

This is great, and I'm sorry for my silence on the subject.. I already have plans for better dynamic queries to be included in v2, I just haven't been able to finish something to show yet.

My idea is to have ``` sql`` ``` work both ways as @Minigugus mentions. Since queries are already deferred to the next tick, any ``` sql`` ``` instance inside another ``` sql`` ``` call will simply not run.

That way I actually think we decrease user facing complexity because we keep the surface api / concepts lower,  and at the same time, automatic syntax highlighting many editors have for ``` sql`` ``` will just work.

It's also better for reusing and composing queries. 

The only caveat is that if you want to reuse or dont want to have your partial query inline it needs to be a function, and not a simple variable assignment. I actually find that a good thing since functions are first class, and many times you might want to pass some parameters as well.

Maybe we could also have an option of doing ``` sql``.partial ``` to support direct variable assignment, just like I'd like to have eg. ``` sql``.query ``` to inspect the actual query sent.

Also related to this, the idea is that ```sql()``` for building queries from objects / arrays will include more features.

yckao:
> Hi @yckao
> 
> This is great, and I'm sorry for my silence on the subject.. I already have plans for better dynamic queries to be included in v2, I just haven't been able to finish something to show yet.
> 
> My idea is to have `` sql`` `` work both ways as @Minigugus mentions. Since queries are already deferred to the next tick, any `` sql`` `` instance inside another `` sql`` `` call will simply not run.
> 
> That way I actually think we decrease user facing complexity because we keep the surface api / concepts lower, and at the same time, automatic syntax highlighting many editors have for `` sql`` `` will just work.
> 
> It's also better for reusing and composing queries.
> 
> The only caveat is that if you want to reuse or dont want to have your partial query inline it needs to be a function, and not a simple variable assignment. I actually find that a good thing since functions are first class, and many times you might want to pass some parameters as well.
> 
> Maybe we could also have an option of doing `` sql``.partial `` to support direct variable assignment, just like I'd like to have eg. `` sql``.query `` to inspect the actual query sent.
> 
> Also related to this, the idea is that `sql()` for building queries from objects / arrays will include more features.

@porsager 
That's sounds good. Thanks for the kindly reply.
Actually I'm also have an idea to make sql`` without promise to work like [knex.toString()](http://knexjs.org/#Interfaces-toString)
So I can achieve what @Minigugus mentions above. Also can store partial query into variables.
(knex do that too.)

Also wondering is there any plan to add some options let user defined their own helpers?

And other problems is the usage of skip. I add this is because we cannot pass undefined into ${} (maybe it can but not in type definition, I haven't find out that).
Maybe we can let user passing undefined into ${} to skip the block?

This project is really great. Maybe you can add some roadmap or tasks, and let's community do it together ?
That's will be awesome if we could pushing amazing feature forward.

stalniy:
I agree with @yckao that `sql` function is over overloaded :) This will make V8 to not be able to optimize this function for faster execution what is a downside.

The idea @yckao presented in https://github.com/porsager/postgres/pull/101#issuecomment-668193835 is much better than relying on  some inernals of event loop

tomm:
I :+1: the original PR. I'd love this feature, since it would allow me to replace Slonik (a similar nodejs postgres interface, which allows sql partials but unfortunately has less than half the performance of porsager/postgres)

Issue #94, opened 2020-07-17
Parsing arrays of custom types


L-Leite:
Hello, thank you for your work developing this library.

I have a project where I fetch a column that stores an array of custom types.
The table structure looks like this:

```sql
CREATE TYPE inventoryItem AS
(
  item_id integer,
  ammount integer
);

CREATE TABLE public.inventories
(
  owner_id integer PRIMARY KEY,
  items inventoryItem[] DEFAULT
  ARRAY[
    /* characters */
    (1005, 1), (1006, 1), (1007, 1) /* etc...*/
   ]::inventoryitem[]
);
```

And I pass these options to postgres:

```ts
import postgres from 'postgres'

export class InventoryItem {
    public item_id: number
    public ammount: number
}

const INVENTORY_ITEM_OID = 12345 /* this is generated by a script */

export type DB_POSTGRES_CUSTOM_TYPES = {
    inventoryItem: InventoryItem
}
export const DB_POSTGRES_CONFIG: postgres.Options<DB_POSTGRES_CUSTOM_TYPES> = {
   /* other options... */
    types: {
        inventoryItem: {
            to: INVENTORY_ITEM_OID,
            from: [INVENTORY_ITEM_OID],
            parse: (str: string): InventoryItem => {
                str = str.substr(1, str.length - 2)
                const parts = str.split(',')

                const id = Number(parts[0])
                const ammount = Number(parts[1])

                return new InventoryItem(id, ammount)
            },
            serialize: (item: InventoryItem): number[] => {
                return [item.item_id, item.ammount]
            }
        }
    }
}

export const sql = postgres(DB_POSTGRES_CONFIG)
```

And I query the data like this:

```ts
export class Inventory {
    public static async getById(userId: number): Promise<Inventory> {
        const resRows = await sql<
            Inventory
        >`SELECT * FROM inventories WHERE owner_id = ${userId};`

        if (resRows.count === 0) {
            return null
        } else if (resRows.count === 1) {
            return resRows[0]
        } else {
            throw new Error('getById: got more than one row for an inventory')
        }
    }
}
```

The issue is that the parse function is never called. I looked into the library's code and I found if I changed [line 245 of index.js](https://github.com/porsager/postgres/blob/806cc2c3a42e453a7fb5292989a74f77a1eeda72/lib/index.js#245) from `where a.typcategory = 'A' and b.typcategory != 'C'` to `where a.typcategory = 'A'` that the parse function gets called.

[...]

PR #103, opened 2020-08-03
Include custom types when fetching array types.

L-Leite:
Removes a 'where' condition in the query used when fetching the arrays types in the database

All tests pass with Node version 14.7.0 and Postgresql 12.3

Fixes issue #94

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -243,7 +243,7 @@ function Postgres(a, b) {
           select b.oid, b.typarray
           from pg_catalog.pg_type a
           left join pg_catalog.pg_type b on b.oid = a.typelem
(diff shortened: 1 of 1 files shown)


porsager:
Thanks @L-Leite 

It's a pity I [didn't write down in detail](https://github.com/porsager/postgres/commit/7d3ce4b19a59ecbbf29f0198a92cc72f208cea0a) why I added the exclusion of composite types in the query. I remember there was a reason for doing so, so I'll just chew on it for a couple of days to see if I come up with why.

porsager:
I'm very sorry this had gotten left behind @L-Leite !

PR #115, opened 2020-10-02
Fix typo, assign result to variable

karlhorky:
Hey @porsager, hope you are well!

Quick PR to fix the missing `:` and assign the result of the query to a variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ module.exports = sql
 // other.js
 const sql = require('./db.js')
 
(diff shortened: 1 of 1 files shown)


porsager:
Right that's probably better since the example is as "in a file". I think I wrote them mimicking the REPL. 

Thanks

karlhorky:
Sounds good, glad to help! Thanks for the merge!

PR #118, opened 2020-10-23
Fix type of timeout

karlhorky:
TypeScript complains that `timeout` needs to be `number | undefined`.

Not sure if `0` is semantically the same as `null` in this case, so happy to fix if not!


porsager:
Ah nice catch. Thanks.

PR #128, opened 2020-11-25
int8 can be represented as number in js

koestler:
The `COUNT(*)` aggregate function in postgres returns a bigint. I think this should be treated as a numeric in javascript.

See this for comments: https://github.com/koestler/postgres/commit/544a7385ff7c9688f7b148b76ed76a23e604431e

However, if a remember correctly, js uses a float64 for all numerics. So not int8 values will have an exact representation.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -10,7 +10,7 @@ const types = module.exports.types = {
   },
   number: {
     to: 0,
-    from: [21, 23, 26, 700, 701],
+    from: [20, 21, 23, 26, 700, 701],
     serialize: x => '' + x,
     parse: x => +x
   },


porsager:
Since it is a bigint you will loose precision unless you cast to js BigInt, but that has it's own issues.

I get that it's not the usual case the count() result is too big to represent in js, but we can't know if the data came from count() or an actual bigint column unless we analyze the query.

Either way, having varying handling depending on the actual result value is not a good thing, so we're pushing the decision to userland.

Nothing prevents you from doing `count(*)::int8`.

If you feel we could improve the docs about it https://github.com/porsager/postgres#numbers-bigint-numeric I'm open to suggestions.

PR #131, opened 2020-12-13
Fix undefined socket if stop postgresql.service at beginning.

szdailei:
1. sudo systemctl stop postgresql.service

2. start program

3. got TypeError: Cannot read property removeListener of undefined at Socket.onclose

Tested on: 
Arch Linux 5.9.13-arch1-1
PostgreSQL 13
postgres 1.0.2 and 2.0.0-beta.2

Btw:
If I start program and then stop postgresql.service, all are OK.
I don't know why got undefined socket if stop postgresql.service at beginning.
I guess nodejs net.connect removes socket.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -259,10 +259,12 @@ function postgresSocket(options, {
   let buffer
 
   function onclose() {
-    socket.removeListener('data', data)
-    socket.removeListener('error', error)
-    socket.removeListener('connect', ready)
-    socket.removeListener('secureConnect', ready)
+    if (socket) {
+      socket.removeListener('data', data)
+      socket.removeListener('error', error)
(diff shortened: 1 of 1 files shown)


porsager:
Hmm.. This smells a little bit. I'd like to dig into why the socket isn't defined when onclose is called first :)

porsager:
I found the smell ;) 

The socket is not being assigned because it's being shadowed in the connect function. Would you mind changing your pull request to remove the `const ` in front of keyword here: https://github.com/porsager/postgres/blob/master/lib/connection.js#L276 ?

porsager:
Fixed in https://github.com/porsager/postgres/commit/c5bfcf2965a8dd65405fdc8ada6fb40071f8deeb

Issue #123, opened 2020-11-13
Authentication error on parallel queries when using scram-sha-256 authentication


marcbachmann:
Hey there, I just tried to write some benchmarks and then found some issue with the `scram-sha-256` authentication.
When running multiple queries in parallel, it will throw an Authentication error.

To reproduce it, you can start two containers, one with md5 authentication, the other with scram-sha-256:
```bash
docker run -d --name postgres-md5 --rm -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:13-alpine
docker run -d --name postgres-scram --rm -p 5433:5432 -e POSTGRES_PASSWORD=password -e POSTGRES_INITDB_ARGS="--auth-host=scram-sha-256 --auth-local=scram-sha-256" postgres:13-alpine
```

Then execute this script:
```js
async function start (port) {
  const sql = require('postgres')(`postgres://postgres:password@localhost:${port}/postgres`, {no_prepare: true})

  try {
    await Promise.all([sql`SELECT 1`, sql`SELECT 1`, sql`SELECT 1`])
  } finally {
    sql.end({timeout: 0})
  }
}

start(5432)
  .catch((err) => console.error('Error from instance with md5 authentication: ', err))

start(5433)
  .catch((err) => console.error('Error from instance with scram authentication: ', err))
```

This results in that error:
```
Error from instance with scram authentication: PostgresError: password authentication failed for user "postgres"
    at Object.ErrorResponse (/benchmark/node_modules/postgres/lib/backend.js:131:22)
    at Socket.data (/benchmark/node_modules/postgres/lib/connection.js:233:25)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:302:12)
    at readableAddChunk (_stream_readable.js:278:9)
    at Socket.Readable.push (_stream_readable.js:217:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23) {
  severity_local: 'FATAL',
  severity: 'FATAL',
  code: '28P01',
  file: 'auth.c',
  line: '331',
  routine: 'auth_failed'
}
```

And sometimes in this one:
```
Error from instance with scram authentication: Error: The server did not return the correct signature
    at Object.SASLFinal (/benchmark/node_modules/postgres/lib/frontend.js:133:16)
    at Object.auth (/benchmark/node_modules/postgres/lib/frontend.js:65:23)
    at /benchmark/node_modules/postgres/lib/connection.js:89:29
[...]

PR #145, opened 2021-01-18
Fix SCRAM-SHA-256 authentication by cloning the options object

marcbachmann:
This pull request fixes the issue for SCRAM authentication mentioned in #123.
The options object is mutated at two places during the authentication to persist connection information. 
To set a `nonce`: https://github.com/porsager/postgres/blob/4d5fdc108a5dac18145c748d13da2fe0d8c02b22/lib/frontend.js#L97
And to set the `serverSignature`: https://github.com/porsager/postgres/blob/4d5fdc108a5dac18145c748d13da2fe0d8c02b22/lib/frontend.js#L122

By cloning the object, we can safely mutate it to successfully do the handshake.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -218,7 +218,9 @@ function Postgres(a, b) {
 
   function createConnection(options) {
     slots--
-    const connection = Connection(options)
+    // The options object gets cloned as the as the authentication in the frontend.js mutates the
+    // options to persist a nonce and signature, which are unique per connection.
+    const connection = Connection({...options})
     all.push(connection)
     return connection
   }
--- a/tests/index.js
+++ b/tests/index.js
@@ -317,6 +317,15 @@ t('Login using scram-sha-256', async() => {
   return [true, (await postgres({ ...options, ...login_scram })`select true as x`)[0].x]
 })
 
+t('Parallel connections using scram-sha-256', async() => {
+  const sql = postgres({ ...options, ...login_scram })
+  return [true, (await Promise.all([
+    sql`select true as x, pg_sleep(0.2)`,
+    sql`select true as x, pg_sleep(0.2)`,
+    sql`select true as x, pg_sleep(0.2)`
+  ]))[0][0].x]
+})
+
 t('Support dynamic password function', async() => {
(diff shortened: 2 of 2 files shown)


lib/index.js
```diff
@@ -218,7 +218,9 @@ function Postgres(a, b) {
 
   function createConnection(options) {
     slots--
-    const connection = Connection(options)
+    // The options object gets cloned as the as the authentication in the frontend.js mutates the
+    // options to persist a nonce and signature, which are unique per connection.
```
marcbachmann, line 222: There's not a single comment in this file. Should I remove it again? 😁

marcbachmann:
Somehow the issue only occured if there are more than two queries running.
With two queries, I didn't get the error.

marcbachmann:
Load tested using that:
```
// proof.js
async function start (port) {
  const sql = require('./')(`postgres://postgres:password@localhost:${port}/postgres`, {max: 100})

  try {
    const promises = []
    for (let index = 0; index < 100000; index++) {
      promises.push(sql`SELECT 1`)
    }
    await Promise.all(promises)
  } finally {
    sql.end({timeout: 0})
  }
}

start(5433)
  .catch((err) => console.error('Error from instance with scram authentication: ', err))
```

mortifia:
load tested with apollo-server and jmeter caling that with no bug

``` js
async coucou() {
        console.log("coucou");

        /*const res = await this.db`
        select *
        from "user".users
        console.log("coucou end");
        return res[0]
        `*/

        const tmp = await Promise.allSettled([
            this.db`select * from "user".users`,
            this.db`select * from "user".users`])
        console.log(tmp)
        console.log("coucou end");

        return res[0].value[0];
    }
```
12 800 request/s with no errors

lib/index.js
porsager: I think that's fine considered the non obvious behavior. Maybe we can look at a fix keeping the nonce internal to the connection later to avoid mutating the options.

porsager:
Thanks a lot @marcbachmann ! Looks good.

Issue #154, opened 2021-02-24
Unexpected return value from 2nd listen call on the same topic


stalniy:
I'm writing tests for `listen` functionality and found that if I add the 2nd listener to the topic that already exists in response I get topic name instead of result with metadata (connection pid, etc):

```js
const result = await sql.listen('test', () => {});
console.log(result);
/*
    [
      count: null,
      command: 'LISTEN',
      state: { status: 'I', pid: 223, secret: -1610071437 }
    ]
*/

const result2 = await sql.listen('test', () => {});
console.log(result2); // test
```

I expect both return the same (or similar) result

stalniy:
by the way, types says that it is always a `PendingRequest` (i.e., `interface PendingRequest extends Promise<[] & ResultMeta<null>> { }`)

porsager:
This was fixed in #155

Issue #133, opened 2020-12-20
Add support for UNLISTEN


patrickmichalina:
Correct me if I am wrong but there does not appear to be a way to call `unlisten` with this library?

joaopaulobdac:
I need a `unlisten` method as well. Did you try using raw sql?
Something like 
```
await sql`UNLISTEN channel`
```
?

Minigugus:
> When you call listen, a dedicated connection will automatically be made to ensure that you receive notifications in real time. This connection will be used for any further calls to listen. Listen returns a promise which resolves once the LISTEN query to Postgres completes, or if there is already a listener active.

You're right, there isn't a proper way to unlisten to a channel. Also, since a dedicated connection is used for listening, `await sql`UNLISTEN channel`;` shouldn't work properly (and if even if it works, it won't remove listeners and cause a memory leak).

The only solution for now seems to create a `sql` instance per listen call, then `sql.end()` on specific instances to "unlisten" I guess :neutral_face:

porsager:
Wow, can't believe I missed this :P Should definitely be added.

stalniy:
@porsager I also need this feature. Looks like feature development is really slow, so I can jump in and do what I care about. So, let's agree on the design and I will send a PR.

My suggestion:

```js
const subscription = await sql.listen('topic', () => {});

subscription.state // { pid: 1, command: 'listen', secret: ... }
await subscription.dispose() // removes callback, in case this is the last callback it calls `unlisten` automatically
```

porsager:
@stalniy yeah, I'm unfortunately really low on time, a PR is much welcomed.

You suggestion looks good, but let's call it `unlisten` instead of dispose.

What is the use case for exposing state?

stalniy:
The state is currently exposed, I thought it's by design :)

porsager:
This is merged and will be included with next beta release

PR #155, opened 2021-02-24
feat: implements unlisten functionality

stalniy:
Relates to #133

BREAKING CHANGE: `listen` function returns an object instead of an empty array. All properties are preserved.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -341,15 +341,18 @@ function Postgres(a, b) {
 
     if (channel in listeners) {
       listeners[channel].push(fn)
-      return Promise.resolve(listener.result)
+      return Promise.resolve(Object.assign({}, listener.result, {
+        unlisten,
+      }))
     }
 
     listeners[channel] = [fn]
     return query({ raw: true }, listener.conn, 'listen ' + escape(channel))
-      .then((response) => {
-        response.unlisten = unlisten
-        listener.result = response
-        return response
+      .then((result) => {
+        listener.result = result
+        return Object.assign({}, listener.result, {
+          unlisten,
+        })
       })
 
     function unlisten() {
@@ -361,7 +364,7 @@ function Postgres(a, b) {
 
       if (!listeners[channel].length) {
         delete listeners[channel]
-        return query({ raw: true }, getListener(), 'unlisten ' + escape(channel))
+        return query({ raw: true }, getListener().conn, 'unlisten ' + escape(channel))
           .then(() => {})
       }
       return Promise.resolve()
--- a/lib/index.js
+++ b/lib/index.js
@@ -344,6 +344,21 @@ function Postgres(a, b) {
 
     listeners[channel] = [fn]
     return query({ raw: true }, getListener(), 'listen ' + escape(channel))
+      .then((result) => {
+        result.unlisten = unlisten
+        return result
+      })
+
+    function unlisten() {
+      const index = listeners[channel].indexOf(fn)
+      if (index !== -1)
+        listeners[channel].splice(index, 1)
+      if (!listeners[channel].length) {
+        delete listeners[channel]
+        return query({ raw: true }, getListener(), 'unlisten ' + escape(channel))
+      }
+      return Promise.resolve()
+    }
   }
 
   function getListener() {
@@ -351,7 +366,7 @@ function Postgres(a, b) {
       return listener
 
     listener = Connection(Object.assign({
-      onnotify: (c, x) => c in listeners && listeners[c].forEach(fn => fn(x)),
+      onnotify: (c, x) => c in listeners && listeners[c].slice(0).forEach(fn => fn(x)),
       onclose: () => {
         Object.entries(listeners).forEach(([channel, fns]) => {
           delete listeners[channel]
--- a/lib/index.js
+++ b/lib/index.js
@@ -353,12 +353,16 @@ function Postgres(a, b) {
       })
 
     function unlisten() {
-      const index = listeners[channel].indexOf(fn)
-      if (index !== -1)
-        listeners[channel].splice(index, 1)
+      if (!listeners[channel]) {
+        return Promise.resolve()
+      }
+
+      listeners[channel] = listeners[channel].filter(handler => handler !== fn)
+
       if (!listeners[channel].length) {
         delete listeners[channel]
         return query({ raw: true }, getListener(), 'unlisten ' + escape(channel))
+          .then(() => {})
       }
       return Promise.resolve()
     }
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -344,14 +344,29 @@ function Postgres(a, b) {
...
+    }
   }
 
   function getListener() {
     if (listener)
       return listener
 
     listener = Connection(Object.assign({
-      onnotify: (c, x) => c in listeners && listeners[c].forEach(fn => fn(x)),
+      onnotify: (c, x) => c in listeners && listeners[c].slice(0).forEach(fn => fn(x)),
```
porsager, line 369: What is the purpose of making a copy of the array here?

porsager (changes requested):
I think it'd be great if we could fix subsequent listen calls not receiving the result, in this PR too. Are you up for that? It effects the way unlisten is implemented, but it's not much use if only the first listener is able to unlisten.

lib/index.js
stalniy: this is standard issue with event buses. I do mutate an array in unlisten. If somebody decides to do this:

```
const s = await sql.listen(“test”, () => s.unlisten());
```

It will mutate an array during iteration, what will cause indexes shift and results in some listeners not being called. I had a similar issue in my lib: https://github.com/stalniy/casl/issues/159

porsager: Ah, of course! That's so subtle, but makes perfect sense. Thanks!

stalniy:
Also Fixes #154

lib/index.js
Minigugus: Why not making a copy of the array in `unlisten`, so that it is copied less often?

`listeners[channel] = listeners[channel].slice(0).splice(index, 1)`

https://github.com/porsager/postgres/pull/155/files#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R357-R358

lib/index.js
```diff
@@ -337,31 +337,51 @@ function Postgres(a, b) {
...
+        listener.result = response
+        return response
+      })
+
+    function unlisten() {
+      const index = listeners[channel].indexOf(fn)
+      if (index !== -1)
+        listeners[channel].splice(index, 1)
+      if (!listeners[channel].length) {
+        delete listeners[channel]
```
Minigugus, line 360: Won't this mean that the next calls to `unlisten` will raise an exception, since after the first call `listeners[channel]` is `undefined`?

Minigugus (commented):
Great work! And don't forget to update the types :wink:

https://github.com/porsager/postgres/blob/master/types/index.d.ts#L340

(I can help with types if you want by the way)

lib/index.js
stalniy: good point!

stalniy: good point!

stalniy: Fixed

stalniy:
updated types

stalniy:
@porsager do you have any feedback after updates?

porsager:
I'll take a look tonight. 

Would you review the types @Minigugus ?

lib/index.js
```diff
@@ -337,31 +337,55 @@ function Postgres(a, b) {
   }
 
   function listen(channel, fn) {
+    const listener = getListener();
+
     if (channel in listeners) {
       listeners[channel].push(fn)
-      return Promise.resolve(channel)
+      return Promise.resolve(listener.result)
```
porsager, line 344: This will result in any listener receiving an unlisten function that only works for the initial listener.

Here's one of the test cases modified to test this.
```js
ot('multiple listeners and unlisten one', async() => {
  const listener = postgres(options)
      , xs = []

  await listener.listen('test', x => xs.push('1', x))
  const s2 = await listener.listen('test', x => xs.push('2', x))
  await listener.notify('test', 'a')
  await delay(50)
  await s2.unlisten()
  await listener.notify('test', 'b')
  await delay(50)
  listener.end();

  return ['1a2a1b', xs.join('')]
})
```

stalniy: Ah, right... unlisten is different for every `fn`

stalniy: fixed

stalniy:
I changed return type of `sql.listen` because it seems strange to copy an array with `Object.assign` but if you would like the return result to be an array, I will do this

Minigugus:
> that array is always empty, so very unlikely that anybody relied on it to be an array

Ok, it's not a bad idea since it hide implementation details and simplify the public API :)

It's also a breaking change, maybe you should also update the changelog :wink:

stalniy:
> Ok, it's not a bad idea since it hide implementation details and simplify the public API :)

where can I find it?

Minigugus:
> where can I find it?

Ok I thought there was a changelog file, my bad x)

Then there is nothing to, only @porsager is responsible for what is in the changelog on the release page.

Minigugus (approved):

porsager:
I think it's fine that the return type is simply `{ unsubscribe }`.

With regards to the changelog, add it as part of the PR description and then I'll add it to releases.

stalniy:
> I think it's fine that the return type is simply `{ unsubscribe }`.
> 
> With regards to the changelog, add it as part of the PR description and then I'll add it to releases.

shall I change the return type?

porsager:
I don't think there's any use case for having the result included, so if you agree let's do it :)

porsager (approved):

stalniy:
@porsager looks like we can't leave `unlisten` only. there is a test that check `listen` reconnect:

```js
t('listen reconnects', async() => {
  const listener = postgres(options)
      , xs = []

  const { state: { pid } } = await listener.listen('test', x => xs.push(x))
  await sql.notify('test', 'a')
  await sql`select pg_terminate_backend(${ pid }::int)`
  await delay(50)
  await sql.notify('test', 'b')
  await delay(50)
  listener.end()

  return ['ab', xs.join('')]
})

```

stalniy:
I actually like that it returns a `pid` because I planned to implement similar set of tests on my side :)

porsager:
Well i guess we're ready then 😊

porsager:
That was a little too soon though. The pid will change if the underlying socket reconnects. Let's tackle that in another PR

stalniy:
Hmm... I think a good way to handle this is to turn `pid` into a function/getter. So, we do not update this value for every subscription

porsager:
Yeah, sounds good.

stalniy:
@porsager how to test unequality or comparison? I'd like to add a test that checks unequality of initial pid and a pid which I get after terminating connection:


```js
t('listen result has state getter', async() => {
  const listener = postgres(options)
      , xs = []

  const result = await listener.listen('test', x => xs.push(x))
  const initialPid = result.state.pid
  await sql.notify('test', 'a')
  await sql`select pg_terminate_backend(${ initialPid }::int)`
  await delay(50)
  listener.end()

  return [result.state.pid, initialPid] // <--- looks like there is no way to check unequality
})
```

PR #158, opened 2021-03-05
Multihost support for High Availability setups

porsager:
Multiple hosts can be specified like in [psql](https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-MULTIPLE-HOSTS).

It's either done through the connection string, or through arrays in the host and port parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ You can use either a `postgres://` url connection string or the options to defin
 
 ```js
 const sql = postgres('postgres://username:password@host:port/database', {
-  host            : '',         // Postgres ip address or domain name
-  port            : 5432,       // Postgres server port
+  host            : '',         // Postgres ip address[s] or domain name[s]
+  port            : 5432,       // Postgres server port[s]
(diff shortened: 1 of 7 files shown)


porsager:
Would you mind taking a look at this @Minigugus and @stalniy ?

Minigugus:
I haven't tested yet but looks good to me. I can try to test it this week if you want, but I never used the postgresql's multi hosts feature, I can't promise anything.

Do you want me to update typescript types by the way?

Minigugus:
I corrected some problems with the types along the way.

I also have an improvement for B but I wanted your opinion before committing, since it may be a breaking change. It improves support for multi-statement queries with `unsafe`. It's not a big deal, it doesn't deserve another PR I think,  but it's off topic for this one. Should I push this anyway?

porsager:
@Minigugus looks great, thank you - are any of the unrelated type changes breaking in any way?

re: `unsafe` Since it's unrelated, let's just take it in another PR

PR #160, opened 2021-03-09
fix(listen): makes sure unlisten returns a reference to conn result

stalniy:
Eventually, this allows to get the new pid of pg connection after `listen` reconnects.

**Follow up PR** for #155
**Issue in tests**: how can I test unequality in tests?

```js
t('listen result reports correct connection state after reconnection', async() => {
  const listener = postgres(options)
      , xs = []

  const result = await listener.listen('test', x => xs.push(x))
  const initialPid = result.state.pid
  await sql.notify('test', 'a')
  await sql`select pg_terminate_backend(${ initialPid }::int)`
  await delay(50)
  listener.end()

  return [result.state.pid, initialPid]
})
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -341,15 +341,19 @@ function Postgres(a, b) {
 
     if (channel in listeners) {
       listeners[channel].push(fn)
-      return Promise.resolve(Object.assign({}, listener.result, { unlisten }))
+      return Promise.resolve(Object.create(listener.result, {
+        unlisten: { value: unlisten }
+      }))
     }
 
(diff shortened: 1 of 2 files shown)


porsager:
Cool @stalniy !

To test inequality you can simply do `[this !== that, true]`

stalniy:
right! So, simple :)

stalniy:
@porsager any plans to merge and release this?

porsager:
Ah, sorry. Didn't see your latest changes. Looks good.

PR #165, opened 2021-03-25
Adding in sql.raw method to allow postgres methods in template values

hobberwickey:
My use case for this was this, which is does't seem like there's a way to do currently. 

```
sql`
    INSERT INTO table (
        id,
        name
    ) VALUES (
        ${ item.id || sql.raw("DEFAULT") },
        ${ item.name }
    )
    ON CONFLICT ON CONSTRAINT table_pkey
    DO UPDATE SET
        name" = ${ item.name }
`
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -513,15 +513,21 @@ function Postgres(a, b) {
   }
 
   function addValue(x, xargs, types) {
-    const type = getType(x)
-        , i = types.push(type.type)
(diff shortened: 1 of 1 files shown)


porsager:
This is tracked in https://github.com/porsager/postgres/issues/12 and should be handled by using tagged template strings to keep Postgres.js safe by default and avoid accidental sql injection.

PR #167, opened 2021-03-29
feat: adds `raw` property to error and debug parameters

### LINKED ISSUE #140: Expose raw values for parameters, before serialization
To illustrate what I mean (silly example that generates error):

```js
sql`
  SELECT CONT(id)::int as amount
  FROM "person"
  WHERE nationality = ${'UK'} and id = ${1}
`
```

This query fails because of typo and the error has `parameters` property. When I check this property, I see this:
 
```js
 [ { type: 0, value: 'UK' }, { type: 0, value: '1' } ]
```

I understand that this is serialized parameters with their data types in Postges but this is not what I passed to `sql` function. So, it's not fine to log something like this into log file (or even in stack trace) because it confuses. I also cannot map values to `value` property because then it looks like I passed string `'1'` to id but I passed a normal number. This becomes even more confusing when we deal with booleans and the value equals to `f` or `t`

`debug` option behaves the same way, I'd like to have a way to get back parameters I passed to `sql` function without writing additional logic on my end to unserialize all of those

stalniy:
Fixes #140

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -525,14 +525,15 @@ function Postgres(a, b) {
 
   function getType(x) {
     if (x == null)
-      return { type: 0, value: x }
+      return { type: 0, value: x, raw: x }
 
     const value = x.type ? x.value : x
         , type = x.type || inferType(value)
 
     return {
       type,
-      value: (options.serializers[type] || types.string.serialize)(value)
+      value: (options.serializers[type] || types.string.serialize)(value),
+      raw: x
     }
   }
 }
--- a/tests/index.js
+++ b/tests/index.js
@@ -1075,9 +1075,14 @@ t('Error contains query string', async() => [
   (await sql`selec 1`.catch(err => err.query))
 ])
 
-t('Error contains query parameters', async() => [
+t('Error contains query serialized parameters', async() => [
   '1',
   (await sql`selec ${ 1 }`.catch(err => err.parameters[0].value))
+])
+
+t('Error contains query raw parameters', async() => [
+  1,
+  (await sql`selec ${ 1 }`.catch(err => err.parameters[0].raw))
 ])
 
 t('Query string is not enumerable', async() => {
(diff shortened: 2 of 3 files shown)


stalniy:
There is a breaking change in `Parameter<T>` type but this is actually not a breaking change but a fix. Because `Parameter<number>` currently is equivalent of `{ type: number, value: number }` but it's **not TRUE** in runtime because `Parameter<number>["value"]` is always `string` (it's a serializable value).

@Minigugus please correct me if I missed something

stalniy:
Also I'd like to get a new beta release when this is merged :) @porsager could you please clarify when you plan to release it?

porsager:
This evening 😊

Minigugus (approved):
@stalniy I just realized that `value` is the result of the `serialize` method from custom types, so you're right, it's always `string`, my bad :confused: 

Your fix looks good to me :+1:

stalniy:
by the way, it's not a string in all cases. [there is a possibility for it to be `null`](https://github.com/porsager/postgres/blob/master/lib/index.js#L528). So, probably better to change it to `string | null`, what are you thoughts guys?

porsager:
Sorry, vacation with children is unpredictable. I'm first at the computer now. Do you want to have the `string | null` part included in this PR before I merge?

stalniy:
Yes. I think it should even be `string | undefined | null`. Because there is instruct equality but I need to test it first. And ensure that this are actually a valid cases in runtime.

stalniy:
`undefined` is not possible in runtime. The library throws an error before running query. That means neither `debug` not query with `parameters` property are created. So, it's safe to type it as `string | null`

stalniy:
updated

porsager:
Exactly 😊 

Looks good. I'll release a new beta tonight

stalniy:
@porsager is there anything I can do to make beta.5 released? I would like to use those new features in my apps :)

porsager:
🙈 Yeah, your reminder just now should do it - sorry I forgot. I'll get it out now.

porsager:
Released ;)

stalniy:
Awesome! Thank you!

PR #176, opened 2021-04-19
feat: adds support for prepared statements in `sql.unsafe`

### LINKED ISSUE #122: Expose possibility to create prepared queries manually
## Description

I have a separate library that converts JSON objects into WHERE condition. This library produces a raw SQL string with placeholders inside (i.e., `name = $1 and active = $2`). So, it prevents possibility of SQL injection. I want to use objects for WHERE conditions because I need to allow clients of my API to send object queries like this:

```js
const query = {
  firstName: 'John',
  age: { lt: 12 }
};
```

To merge `postgres` with that library, I use `sql.unsafe` which is safe for my use cases. But **what I miss is the possibility to create prepared statements for this dynamic queries** (I have no more than 10th of them and usually 3-5 are used very often)

## Desired behavior

Please, expose possibility to make prepared queries manually.

**P.S.**: thanks for the awesome library, I really like it!

stalniy:
Fixes #122

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -67,7 +67,7 @@ function Postgres(a, b) {
     , listener
 
   function postgres(xs) {
-    return query({}, getConnection(), xs, Array.from(arguments).slice(1))
+    return query({ prepare: options.prepare }, getConnection(), xs, Array.from(arguments).slice(1))
   }
 
   Object.assign(postgres, {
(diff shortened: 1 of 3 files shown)


porsager:
That looks great @stalniy - Thanks!

porsager:
That looks great @stalniy - Thanks!

stalniy:
I think that I forgot to update README with new `prepare: true/false` flag and possibility to make unsafe prepare queries

PR #184, opened 2021-05-23
fix(cursor): make sure the last batch of rows is not lost - fixes #150, #81

### LINKED ISSUE #81: Cursor custom n does not work for n > nb rows
This test pass:

```js
t('Cursor custom n works', async() => {
  const order = []
  await sql`select * from generate_series(1,20)`.cursor(10, async(x) => {
    order.push(x.length)
  })
  return ['10,10', order.join(',')] // => ['10', '10']
})
```

This test fails:

```js
t('Cursor custom n works', async() => {
  const order = []
  await sql`select * from generate_series(1,20)`.cursor(11, async(x) => {
    order.push(x.length)
  })
  return ['11,9', order.join(',')] // => ['11,9', '11']
})
```

This test pass:

```js
t('Cursor custom n works', async() => {
  const order = []
  const result = await sql`select * from generate_series(1,20)`.cursor(11, async(x) => {
    order.push(x.length)
  })
  return ['11,9', order.concat(result.length).join(',')] // => ['11,9', '11,9']
})
```

Possible workarround:

```diff
--- postgres/lib/connection.js
+++ postgres/lib/connection.js
@@ -196,7 +196,15 @@
         error(err)
       }
     } else if (backend.query) {
-      backend.query.resolve(backend.query.results || backend.query.result)
+      const query = backend.query
+      if (query.cursor && query.result.length)
+        new Promise(r => r(query.cursor(
+          query.cursor.rows === 1 ? query.result[0] : query.result
+        )))
+          .then(() => query.resolve(query.results || query.result))
+          .catch(err => query.reject(err))
+      else
+        query.resolve(query.results || query.result)
     }
 
     backend.query = backend.error = null
```

### LINKED ISSUE #150: Final rows lost with batch mode of PendingQuery.cursor
If the number of results isn't divisible by the batch size, the last (and incomplete) batch is never processed:

```js
const p = require("postgres");
const q = p("postgresql://postgres:example@localhost/demo");

// this works and prints all 9 rows
console.log("a:")
const a = q`select * from "aircrafts"`.cursor(console.log)

setTimeout(() => {
  // this only prinst the first 5 rows
  console.log("b:")
  const b = q`select * from "aircrafts"`.cursor(5, console.log);
}, 1000);

setTimeout(() => {
  // this only prinst the first 8 rows
  console.log("c:")
  const c = q`select * from "aircrafts"`.cursor(2, console.log);
}, 2000);

setTimeout(q.end, 3000)
```

<details><summary>Setting up</summary>

```yml
# docker-compose.yml
# Use postgres/example user/password credentials
version: '3.1'

services:

  db:
    image: postgres:9
    restart: always
    environment:
      POSTGRES_PASSWORD: example
    volumes:
      - ${PWD}/data:/var/lib/postgresql/data
    ports:
      - 5432:5432
```

- Start postgres

```
mkdir -p ~/postgres/data
# copy above docker-compose.yml to ~/postgres/docker-compose.yml
cd ~/postgres;
docker-compose up -d;
```

- Add a database called `demo` to postgres

```
docker-bash
# select the postgres db container
# install some dependencies
apt update
apt install curl unzip -y
# download a test db dump
curl https://edu.postgrespro.com/demo-small-en.zip > demo.zip
unzip demo.zip
# restore the test db
PGPASSWORD=example psql --username=postgres  -f demo-small-en-20170815.sql
```
</details>

naumf:
Fixes #150, fixes #81.
There is no portal suspended indicator in these use-cases as it only appears if an Execute message's row-count limit was reached.
In these cases it is not reached since returned rows are less than the row-count limit, so the cursor's callback is called with the last batch of rows on command complete.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -75,18 +75,15 @@ function Connection(options = {}) {
   }
 
   function oncomplete() {
-    if (backend.query.cursor) {
-      const length = backend.query.result.length
-      if (length && length < backend.query.cursor.rows) {
-        new Promise(r => r(backend.query.cursor(
-          backend.query.cursor.rows === 1 ? backend.query.result[0] : backend.query.result
-        ))).then(() => socket.write(frontend.Close())).catch(err => {
-          backend.query.reject(err)
-          socket.write(frontend.Close())
-        })
-      } else {
+    if (backend.query.cursor && backend.query.result.length) {
+      new Promise(r => r(backend.query.cursor(
+        backend.query.cursor.rows === 1 ? backend.query.result[0] : backend.query.result
+      ))).then(() => socket.write(frontend.Close())).catch(err => {
+        backend.query.reject(err)
         socket.write(frontend.Close())
-      }
+      })
+    } else if (backend.query.cursor) {
+      socket.write(frontend.Close())
     }
   }
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -75,7 +75,19 @@ function Connection(options = {}) {
   }
 
   function oncomplete() {
-    backend.query.cursor && socket.write(frontend.Close())
+    if (backend.query.cursor) {
+      const length = backend.query.result.length
+      if (length && length < backend.query.cursor.rows) {
+        new Promise(r => r(backend.query.cursor(
+          backend.query.cursor.rows === 1 ? backend.query.result[0] : backend.query.result
+        ))).then(() => socket.write(frontend.Close())).catch(err => {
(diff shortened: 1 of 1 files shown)


porsager:
Thanks a lot for the PR @naumf , I finally had time to look at this, and wanted to be sure I understood the issue properly myself, so I ended up implementing a fix myself in https://github.com/porsager/postgres/commit/cea5f73121ca45a346c5118c3a9cb907717d8473 .

PR #205, opened 2021-07-28
Optionally disable fetching of array types.

### LINKED ISSUE #202: `origin` property missing on query in `fetchArrayTypes` query
This isn't really postgres.js fault this is my fault, but I wanted to report it just incase it was helpful, as an unrelated error obscures the actual error and it raises a few interesting questions.

I revoked all permissions from `public` so that new roles would not have the ability to query anything that wasn't explicitly granted.  However this breaks the `fetchArrayTypes` call as this line ` err.stack += backend.query.origin.replace(/.*\n/, '\n')` crashes as `backend.query.origin` is `undefined`.  I think this is because the query object is manually constructed.

To repro:

1. Remove all grants from all tables in the catalog and info schema:

```sql
revoke all on all tables in schema pg_catalog from public;
revoke all on all tables in schema information_schema from public;				
```

2. Create a new role 

3. Connect as that role with postgres.js and run a simple query e.g. `select 1`

You should get an unhandled exception "cannot call replace on undefined"

---

On a related note, is there a recommended way to work around fetching the array types?  I'm trying to ensure a database role has only has access to what was explicitly granted.  I'm guessing I'm missing something very basic.  I probably missed something in the readme again 🙈.

Alternatively, I could just grant what is used by that query, but it would be great to know the background there.

JAForbes:
An alternative or complement to #204

Fixes #202

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ const sql = postgres('postgres://username:password@host:port/database', {
   },
   target_session_attrs : null   // Use 'read-write' with multiple hosts to 
                                 // ensure only connecting to primary
+  fetch_array_types    : true   // Disable automatically fetching array types
+                                // on initial connection.
 })
 ```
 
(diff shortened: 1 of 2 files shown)


porsager:
This looks just fine 👍 Would you mind adding the option to the README.md?

JAForbes:
Done

porsager:
Awesome - thanks!

Issue #163, opened 2021-03-23
Syntax to release a connection rather than kill the pool?


kingram6865:
I don't see the command to release a connection once a db call is completed.
is <>.end() the only option? That doesn't make sense. When I use that one I can't make any other db calls. The pool is gone.

porsager:
There is no manual handling of connections, unless you create a separate postgres instance with `max: 1` like `postgres({ max: 1 })` and call `.end()` on that.

You can also set `idle_timeout` so that connections are closed automatically after a set period of inactivity.

porsager:
Feel free to reopen if you think there's a better way to achieve this or if I misunderstood your question 🙂

kingram6865:
> Feel free to reopen if you think there's a better way to achieve this or if I misunderstood your question

You got it. I hear, understand and comprehend.

PR #206, opened 2021-07-30
Connection lifetime

metala:
Rationale.
In order to mitigate the issue of long running connections that increase database memory usage over time I have introduced lifetime to all connections of the pool. I need your feedback on the patch.


iby:
@metala Hey there! Just curious if this attempts to solve #163 and removes hanging connections after completed queries?

metala:
@iby Actually, it's about an OOM issue I hit with PostgreSQL as one of the services was making a lot of specific queries and crashing the database server. It might work for that case as well, but the connection lifetime is set using time duration and not number of requests.

The issue is about how Linux's fork() and memory CoW work. It is better explained here: https://italux.medium.com/postgresql-out-of-memory-3fc1105446d

porsager:
Yeah, #163 didn't really have an issue did it? what are you having issues with @iby ? (maybe open a new issue)

metala:
I am trying to run all the tests, but the test runner is broken on my setup. I have just ensured that my new tests are running fine.

My test setup:
```
$ docker run -it -e 'POSTGRES_USER=metala' -e 'POSTGRES_DATABASE=postgres_js_test' -e 'POSTGRES_HOST_AUTH_METHOD=trust' -v '/var/run/postgresql/:/var/run/postgresql/' postgres:13
$ yarn test
```

lib/index.js
```diff
@@ -228,7 +228,20 @@ function Postgres(a, b) {
...
+    let connection
+    while (1) {
+      connection = slots ? createConnection(options) : connections.shift()
+      if (!connection || !connection.hasExpired()) {
+        break
+      }
+
+      // has expired
+      connection.release()
+      const idx = all.findIndex(x => x === connection)
```
metala, line 240: there is a bug here. `connection` is in a closure and the variable is outside the loop.

metala:
We are switching DB, so I don't think I would be able to maintain my change here, thus I will just close the MR.

PR #212, opened 2021-08-13
transform.column.to support

porsager:


iby:
@porsager Howdy! Could you beta-release this please?

PR #227, opened 2021-09-08
Expose PostgresError on module

### LINKED ISSUE #226: Make PostgresError available in module
I need to access `PostgresError` to run violation checks. Right now it's declared like below along with case transformers.

```ts
declare namespace postgres {
  type PostgresError = PostgresErrorType
  function toPascal(str: string): string;
  function toCamel(str: string): string;
⠇
```

Case transformers are directly available in the import, but `PostgresError` isn't. It's only available via an existing sql connection, which is really inconvenient to access in a helper file where it doesn't exist.

```ts
console.log(postgres.toCamel);                  // [Function (anonymous)]
console.log(postgres.prototype.PostgresError);  // undefined
console.log(error instanceof sql.PostgresError); // true
```

Can `PostgresError` be exposed directly and made easily accessible in the module? Perhaps there are other handy types in the same situation?

iby:
Fixes #226. @porsager I think it would be neater to just move `PostgresError` declaration into `postgres` namespace to avoid that intermediary `PostgresErrorType` and `const`. Didn't want to make too big of a change, so just went with that. Let me know if you want it in the namespace and I'll amend! 👍

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -33,6 +33,7 @@ function notTagged() {
 }
 
 Object.assign(Postgres, {
+  PostgresError,
   toPascal,
   toCamel,
   toKebab,
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -103,10 +103,10 @@ type UnwrapPromiseArray<T> = T extends any[] ? {
   [k in keyof T]: T[k] extends Promise<infer R> ? R : T[k]
 } : T;
 
-type PostgresErrorType = PostgresError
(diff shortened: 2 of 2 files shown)


iby:
@porsager Also, it would be 🥳 if you can do Beta 9 after this. Eager to get these two-way transformers into the project.

porsager:
Awesome - thanks @iby 

I'll make a release today :)

porsager:
Hey @iby .. I completely forgot about making that release - sorry.. It's done now though.

PR #229, opened 2021-09-09
WIP: fix slow parsing of large messages

porsager:
Following the discussion in https://github.com/porsager/postgres/pull/207 I made this fix. I might improve the code a little before merging, but this introduces only performance improvements and no performance regressions.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -295,7 +295,7 @@ function Connection(options = {}) {
     }
 
     buffer = buffers
-      ? Buffer.concat(buffers)
+      ? Buffer.concat(buffers, length - remaining)
       : buffer.length === 0
         ? x
         : Buffer.concat([buffer, x], buffer.length + x.length)
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -33,6 +33,8 @@ function Connection(options = {}) {
   let next = false
   let statements = {}
   let connect_timer
+  let bs = null
+  let rest = 0
 
   const queries = Queue()
       , id = count++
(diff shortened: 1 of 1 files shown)


eugene1g:
The logic is philosophically the same as alt solutions in that PR 👍

lib/connection.js
```diff
@@ -285,17 +287,31 @@ function Connection(options = {}) {
...
+      rest -= x.length
+      if (rest >= 0)
+        return
+    }
+
+    buffer = bs
+      ? Buffer.concat(bs)
+      : buffer.length === 0
+        ? x
+        : Buffer.concat([buffer, x], buffer.length + x.length)
```
eugene1g, line 301: Defining the length of the new Buffer (via `buffer.length + x.length`) is not necessary unless you want the new buffer to be larger/smaller than `buffer.length + x.length`

lib/connection.js
```diff
@@ -33,6 +33,8 @@ function Connection(options = {}) {
   let next = false
   let statements = {}
   let connect_timer
+  let bs = null
```
eugene1g, line 36: Maybe naming this as `bufferStack` or `bufferQueue` would help to understand what's going on 6 months later :D

stalniy:
have no idea how it works :D but it's awesome that it improves the performance for big chunks :)

lib/connection.js
porsager: I recall early on I did some perf testing on this, and it's faster because node will internally iterate the array to get the length.

lib/connection.js
porsager: Good idea 😉 But `bs` is such a nice variable name 🤪

PR #233, opened 2021-09-22
Support Deno

porsager:
After doing the [esm conversion](https://github.com/porsager/postgres/pull/231) I couldn't help but try and see how far away Deno support was.

This was done by using the deno std/node modules and some polyfills for the missing parts. It's using unstable parts, so needs to run with `--unstable` for now.

All tests are passing except for 3 which used `fs.createReadStream` and `net.createServer`.

This of course needs a lot more thought, but if you're using Deno and want to play around with it I'd love to know if it's any good.

## Quick start

```js
// yay.js
import postgres from 'https://raw.githubusercontent.com/porsager/postgres/deno/deno/lib/index.js'

const sql = postgres()

console.log(
  await sql`select 1`
)
```

```sh
deno run --unstable --allow-env --allow-net wat.js
```


TriForMine:
Hey, this is really awesome that the lib now supports deno.
Deno database drivers are really bad and have a lot of issues. And so far using this library is way much more stable and faster than any alternative avalaible.
I was using a fork that was made for deno, but that branch is even faster.

Thanks so much for the hard work on that lib!

Ps: If you want types with deno you have to import it like that:
```ts
// @deno-types="https://raw.githubusercontent.com/porsager/postgres/deno/types/index.d.ts"
import postgres from 'https://raw.githubusercontent.com/porsager/postgres/deno/deno/lib/index.js'
```

porsager:
Hi @TriForMine .. Thank you! Glad it's of use ☺️

And thank you for the typescript note!

Once the esm branch is merged I guess we can merge this too and add some info about "experimental deno" support in the readme. If you bump into any issues be sure to let me know.

PR #235, opened 2021-09-29
Update transformer types and documentation

iby:
Compliments #212.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,34 +45,34 @@ You can use either a `postgres://` url connection string or the options to defin
 
 ```js
 const sql = postgres('postgres://username:password@host:port/database', {
-  host                 : '',            // Postgres ip address[s] or domain name[s]
-  port                 : 5432,          // Postgres server port[s]
-  path                 : '',            // unix socket path (usually '/tmp')
-  database             : '',            // Name of database to connect to
-  username             : '',            // Username of database user
-  password             : '',            // Password of database user
-  ssl                  : false,         // true, prefer, require, tls.connect options
-  max                  : 10,            // Max number of connections
-  idle_timeout         : 0,             // Idle connection timeout in seconds
-  connect_timeout      : 30,            // Connect timeout in seconds
-  no_prepare           : false,         // No automatic creation of prepared statements
-  types                : [],            // Array of custom types, see more below
-  onnotice             : fn,            // Defaults to console.log
-  onparameter          : fn,            // (key, value) when server param change
-  debug                : fn,            // Is called with (connection, query, params)
+  host                 : '',                   // Postgres ip address[s] or domain name[s]
+  port                 : 5432,                 // Postgres server port[s]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,34 +45,34 @@ You can use either a `postgres://` url connection string or the options to defin
...
+  onnotice             : fn,                   // Defaults to console.log
+  onparameter          : fn,                   // (key, value) when server param change
+  debug                : fn,                   // Is called with (connection, query, params)
   transform            : {
-    column             : fn,            // Transforms incoming column names
-    value              : fn,            // Transforms incoming row values
-    row                : fn             // Transforms entire rows
+    column             : {to?: fn, from?: fn}, // Transforms incoming or outgoing column names
+    value              : {to?: fn, from?: fn}, // Transforms incoming or outgoing row values
+    row                : {to?: fn, from?: fn}, // Transforms entire incoming or outgoing rows
```
iby, line 66: @porsager wasn't sure of the best way to represent this. Please tweak if have a better idea.

types/index.d.ts
```diff
@@ -52,12 +52,12 @@ interface BaseOptions<T extends JSToPostgresTypeMap> {
...
-    /** Transforms incoming row values */
-    value?: (value: any) => any;
-    /** Transforms entire rows */
-    row?: (row: postgres.Row) => any;
+    /** Transforms incoming and outgoing column names. */
+    column?: (column: string) => string | { from?: (column: string) => string, to?: (column: string) => string };
+    /** Transforms incoming and outgoing row values. */
+    value?: (value: any) => any | { from?: (value: any) => any, to?: (value: any) => any };
+    /** Transforms entire incoming and outgoing rows. */
+    row?: (row: postgres.Row) => any | { from?: (row: postgres.Row) => any, to?: (row: postgres.Row) => any };
```
iby, line 60: @porsager is this the right signature for `to`?

Minigugus: I assumes `from`'s output should be `to`'s input and vice versa, then I would suggest `{ from?: (row: postgres.Row) => any, to?: (row: any) => postgres.Row }`, but I'm not sure where `row.to` could be used...

Also, I guess `( )` are missing in the union? TS sees `(row: postgres.Row) => (any | { from?: (row: postgres.Row) => any, to?: (row: any) => any })`

```suggestion
    row?: ((row: postgres.Row) => any) | { from?: (row: postgres.Row) => any, to?: (row: any) => any };
```

iby: @Minigugus nicely spotted, thank you! Yes, I had the same feeling about the row. I updated my commit with your suggestions to avoid these micro-changes. Please feel free to review further.

porsager:
@Minigugus is doing a big type update here https://github.com/porsager/postgres/pull/257, so closing this for now, but please comment on that PR if this fix is missing.

PR #255, opened 2021-12-28
remove time without time zone from date type cast

dilan-dio4:
Time types are stored as strings that are not compatible in the `new Date` constructer. They don't need to be cast.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -28,7 +28,7 @@ const types = module.exports.types = {
   },
   date: {
     to: 1184,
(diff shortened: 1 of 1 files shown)


porsager:
could you show an example where they are not compatible?

porsager:
Haha, sorry, I read the topic as timestamp without time zone and not time. My bad. You're absolutely right, trying to cast `time` to js `Date` doesn't make any sense.

Thanks a lot for the PR!

dilan-dio4:
Thanks! @porsager

PR #257, opened 2022-01-10
Fix and update types

Minigugus:
TypeScript definitions are missing a lot a features (mostly `.raw()`, `.readable()`, `.writable()`, `sql.subscribe()` and some new options)

I tried to fix them before the v2 release. I also added default values for options in their descriptions, it can be useful for options that fallback to an environment variable when missing. It might be cool to add more documentation too but I'm busy these days :confused:

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -16,37 +16,55 @@ declare function postgres<T extends JSToPostgresTypeMap>(url: string, options?: 
  * Connection options of Postgres.
  */
 interface BaseOptions<T extends JSToPostgresTypeMap> {
-  /** Postgres ip address or domain name */
+  /** Postgres ip address[s] or domain name[s] */
   host: string | string[];
-  /** Postgres server port */
+  /** Postgres server[s] port[s] */
   port: number | number[];
-  /** Name of database to connect to */
+  /** unix socket path (usually '/tmp') */
+  path: string | undefined;
+  /**
+   * Name of database to connect to
+   * @default process.env['PGDATABASE'] || options.user
+   */
   database: string;
-  /** Username of database user */
+  /**
+   * Username of database user
+   * @default process.env['PGUSERNAME'] || process.env['PGUSER'] || require('os').userInfo().username
+   */
   user: string;
-  /** True; or options for tls.connect */
-  ssl: 'require' | 'prefer' | boolean | object;
-  /** Max number of connections */
+  /**
+   * true, prefer, require or tls.connect options
+   * @default false
+  */
+  ssl: 'require' | 'allow' | 'prefer' | boolean | object;
+  /**
+   * Max number of connections
+   * @default 10
+   */
   max: number;
-  /** Idle connection timeout in seconds */
+  /**
+   * Idle connection timeout in seconds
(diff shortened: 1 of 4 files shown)


porsager:
Hi @Minigugus .. Great to see you back. All those things needs to have documentation written too, so that's also a thing needed before release, but just as you, I'm also quite busy.

I'm done with a pretty big (complete) overhaul of the internals which I'll push soon, and I'm not really sure if it makes sense to bump to v3 instead. It includes the missing dynamic query function, so I think the only thing missing now for a release is getting docs and types up to date.

Minigugus:
@porsager Nice :+1: Can't wait for dynamic queries :laughing:

Since I'm busy, maybe these changes could be merged after a quick review so that anyone wanting to contribute to the documentation can do so?

porsager:
Would you mind merging against the rewrite branch? 😊

types/index.d.ts
```diff
@@ -261,7 +415,8 @@ declare namespace postgres {
     | Helper<any>
     | Parameter<any>
     | ArrayParameter
-    | SerializableParameter[];
+    | Record<string, any> // implicit JSON
```
Minigugus, line 418: I noticed new tests with an interesting name, I guess it's a new feature?

https://github.com/porsager/postgres/blob/43d78d548694dea7643b98c434f139ca4c18fdbf/tests/index.js#L113-L121

porsager: Ah yes.. It's because we're now using the ParameterDescription from postgres to decide on types :)

types/index.d.ts
```diff
@@ -58,15 +76,49 @@ interface BaseOptions<T extends JSToPostgresTypeMap> {
...
+      // to?: (row: postgres.Row) => any; // unused
+    };
   };
   /** Connection parameters */
   connection: Partial<postgres.ConnectionParameters>;
+  /**
+   * Use 'read-write' with multiple hosts to ensure only connecting to primary
+   * @default process.env['PGTARGETSESSIONATTRS']
+   */
+  target_session_attrs: undefined | 'read-write';
```
porsager: could also be one of `'read-only' | 'primary' | 'standby' | 'prefer-standby'`

porsager (commented):
Looks great @minigugus - let me know when you think it's ready to merge and we'll go..

types/index.d.ts
Minigugus: That's weird, it's already fixed... :thinking: Here is the in latest commit (43d78d5):

https://github.com/porsager/postgres/blob/43d78d548694dea7643b98c434f139ca4c18fdbf/types/index.d.ts#L101

Your comment just disappeared from the Files changed menu, maybe it's a Github cache issue? :smile:

porsager: Haha.. I think I might have gone to the previous commit and commented there, and so I didn't see you'd already fixed it 🙈

Minigugus:
@porsager I would say it's ready. Some flaws may remain but most common use cases are covered and I don't have time to chase them and I don't want the release to be delayed again :slightly_smiling_face:

porsager:
Very cool, and understandable! Thanks @Minigugus !

PR #180, opened 2021-04-28
Add note to idle_timeout about ECONNRESET

karlhorky:
Ref: https://github.com/porsager/postgres/issues/179#issuecomment-828387109

PR #259, opened 2022-01-11
v3

### LINKED ISSUE #12: Better dynamic query building

### LINKED ISSUE #30: Feature: Implement [Symbol.asyncIterator] for .cursor() without a function.
Implementation of `{ [Symbol.asyncIterator]: ... }` when calling cursor without a callback function in order to support _for-await-of_ syntax

See Comment by @porsager in https://github.com/porsager/postgres/issues/1#issuecomment-581777536

### LINKED ISSUE #63: Using Postgres functions, like NOW(), in query helpers
Consider the code:

```js
const fields = {
  name: 'John Doe',
  created_at: ...
};
await sql`
  INSERT INTO users ${sql(fields)}
`;
```

What would I use for `created_at` to send `NOW()`?

I know I could probably use the following, but my hope is to keep the SQL compact with the dynamic query helper:

```js
await sql`
  INSERT INTO users (
    name,
    created_at
  ) VALUES (
    ${fields.name},
    NOW()
  )
`;
```

### LINKED ISSUE #65: Multiple cursors on single transaction not supported.
```
  await sql.begin((tx:any)=>{
    return Promise.all([
      tx`select 1 as cursor, x from generate_series(1,4) as x`.cursor(async (row: any[])=>{
        console.log('cursor1' , JSON.stringify(row));
        await new Promise(resolve=>setTimeout(resolve, 2000));
      }),
      tx`select 2 as cursor, x from generate_series(101,104) as x`.cursor(async (row: any[])=>{
        console.log('cursor2' , JSON.stringify(row));
        await new Promise(resolve=>setTimeout(resolve, 1000));
      })
    ]);
  });
```
This code returns unexpected console output: 
```
cursor1 {"cursor":1,"x":1}
cursor1 {"cursor":2,"x":101}
cursor1 {"cursor":2,"x":102}
cursor1 {"cursor":2,"x":103}
cursor1 {"cursor":2,"x":104}
```
is it possible to support simultaneous reading for multiple cursors in single transaction?

### LINKED ISSUE #67: Support large objects
Compare with https://github.com/Joris-van-der-Wel/node-pg-large-object

### LINKED ISSUE #89: Convert to ES6 module
why using module.exports whne you can import/export?

### LINKED ISSUE #156: Allow using postgres.js for templating only
I have noticed that most, if not all, js sql libraries don't allow the use of their templating DSL separate from a DB connection. I for example want to use your templating for generating queries that I might execute at a later moment.

Maybe that's already possible but not documented.

### LINKED ISSUE #159: Expose db socket disconnected/connected events
## Motivation

I'd like to know when db connection goes down and gets up at application level. So, I can put my app in emergency mode. For example, if db connection is not available for 5 mins, I could close the app (`process.exit`) or rely on app caches only and return some additional information for my API clients in response.

## Proposed implementation

I propose to add 2 callback functions `onconnected` and `ondisconnected` (maybe there is a better way to express this through a single function, like `connectionStateChanged` but then it will require to call this function when connection becomes idle, so not sure)

### LINKED ISSUE #179: ECONNRESET errors after a while
Hi @porsager !

Similar to #43, I'm also encountering similar `ECONNRESET` errors with a PostgreSQL service on Render.com, but with a very low number of connections:

```
Apr 27 10:22:53 PM  node:internal/process/promises:246
Apr 27 10:22:53 PM            triggerUncaughtException(err, true /* fromPromise */);
Apr 27 10:22:53 PM            ^
Apr 27 10:22:53 PM  
Apr 27 10:22:53 PM  Error: read ECONNRESET    at TCP.onStreamRead (node:internal/stream_base_commons:211:20) {
Apr 27 10:22:53 PM    errno: -104,
Apr 27 10:22:53 PM    code: 'ECONNRESET',
Apr 27 10:22:53 PM    syscall: 'read'
Apr 27 10:22:53 PM  }
Apr 27 10:22:54 PM  error Command failed with exit code 1.
```

It seems like after a certain amount of time, the connection with PostgreSQL goes into this state where the `ECONNRESET` errors start happening. Once it's in that state, it fails all queries for a while (a few minutes?). But then it eventually "wakes up" again (I guess Postgres.js eventually figured it out and reconnected at that point).

I'm using `postgres@2.0.0-beta.5` - I'm assuming this beta version is not super unstable like this...? I never had any issues with this with Heroku's PostgreSQL service.

I've also asked for assistance from the Render.com support team over here: https://community.render.com/t/econnreset-with-node-js-and-postgresql/1498

### LINKED ISSUE #201: sql.savepoint as tagged template function?
I imagine `savepoint` will often be used for a single query.  Would it make sense to offer sql.savepoint as both a callback and a tagged template function?

E.g.

```js
let unavailable;
sql.savepoint`insert into users(username) ${username}`.catch( () => unavailable = true )
```

vs

```js
let unavailable;
sql.savepoint( sql => sql`insert into users(username) ${username}` ).catch( () => unavailable = true )
```

It might not be worth the sugar, but I thought I'd mention the idea because all my savepoint usages so far had been for a single query.

### LINKED ISSUE #221: Support ParameterDescription (statement variant)
Hello.

The documentation says:
> The Describe message (statement variant) specifies the name of an existing prepared statement (or an empty string for the unnamed prepared statement). The response is a ParameterDescription message describing the parameters needed by the statement
https://www.postgresql.org/docs/13/protocol-flow.html

This means that you will get a list of the types of all input parameters.
Then you just need to encode all the input parameters in accordance with the types that you received from this response.

Then this query will work correctly:
```JS
sql`SELECT ${[1, 2, 3]}::int[]`
```

### LINKED ISSUE #230: Documentation for copy from/to
If anyone would like to contribute some documentation for the copy from/to streams features added, I'd be very grateful. 

Maybe you wanna take a look at it @uasan ?

### LINKED ISSUE #234: Support Canceling Requests
Hello.
I would like to suggest to implement Canceling Requests in Progress, it seems not difficult and often many people need
> During the processing of a query, the frontend might request cancellation of the query. The cancel request is not sent directly on the open connection to the backend for reasons of implementation efficiency: we don't want to have the backend constantly checking for new input from the frontend during query processing. Cancel requests should be relatively infrequent, so we make them slightly cumbersome in order to avoid a penalty in the normal case.
> 
> To issue a cancel request, the frontend opens a new connection to the server and sends a CancelRequest message, rather than the StartupMessage message that would ordinarily be sent across a new connection. The server will process this request and then close the connection. For security reasons, no direct reply is made to the cancel request message.
> 
> A CancelRequest message will be ignored unless it contains the same key data (PID and secret key) passed to the frontend during connection start-up. If the request matches the PID and secret key for a currently executing backend, the processing of the current query is aborted. (In the existing implementation, this is done by sending a special signal to the backend process that is processing the query.)
> 
> The cancellation signal might or might not have any effect — for example, if it arrives after the backend has finished processing the query, then it will have no effect. If the cancellation is effective, it results in the current command being terminated early with an error message.
> 
> The upshot of all this is that for reasons of both security and efficiency, the frontend has no direct way to tell whether a cancel request has succeeded. It must continue to wait for the backend to respond to the query. Issuing a cancel simply improves the odds that the current query will finish soon, and improves the odds that it will fail with an error message instead of succeeding.
> 
> Since the cancel request is sent across a new connection to the server and not across the regular frontend/backend communication link, it is possible for the cancel request to be issued by any process, not just the frontend whose query is to be canceled. This might provide additional flexibility when building multiple-process applications. It also introduces a security risk, in that unauthorized persons might try to cancel queries. The security risk is addressed by requiring a dynamically generated secret key to be supplied in cancel requests.

https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.5.7.9
Thanks

### LINKED ISSUE #248: How to use postgres' "values" and postgres.js to respond with multiple rows
So this works:
```
const db = require('postgres')(/* ... */)
```

```
> db`values (${db([1, 2])})`.then(console.log)
8 values (1,2) [ dynamic: true ]
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 5060,
  [Symbol(trigger_async_id_symbol)]: 5055,
  [Symbol(destroyed)]: { destroyed: false }
}
> [ { column1: 1, column2: 2 }, count: 1, command: 'SELECT' ]
```

That seems to work well. Now I'll try to have the query respond with multiple rows. I'll use the multiple insert in the docs as an example.

```
> db`values (${db([[1, 2], [2, 3]])})`.then(console.log)
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 5377,
  [Symbol(trigger_async_id_symbol)]: 5376,
  [Symbol(destroyed)]: { destroyed: false }
}
> Uncaught TypeError: str[i].charCodeAt is not a function
    at escape (/home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/types.js:78:16)
    at /home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/index.js:416:59
    at Array.reduce (<anonymous>)
    at escapeHelper (/home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/index.js:416:15)
    at parseHelper (/home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/index.js:386:12)
    at parse (/home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/index.js:363:11)
    at send (/home/ant/projects/rezerv/packages/database/node_modules/postgres/lib/index.js:188:65)
```

This is a simple example. I ran into the issue when trying to do the following:

```sql
          insert into users
            select * (
              select *, row_number() over (order by (select 0)) as n
              from accounts;
              where
                email = ${email}
              limit ${users.length}
            ) a join (
              select *, row_number() over (order by (select 0)) as n
              from (values (${sql(users)}))
            ) b
            on (a.n = b.n)`
```

So why would I want to do this? I want to perform a single query with the least amount of back and forth between the server and db. I don't want that network penalty or race conditions if it isn't necessary. In this case it isn't necessary to first perform a select and then perform inserts.

I have also tried hacking my way around but no success.

```
> db`select * from ${db([db.json(1)])}`.then(console.log)
8 select * from [object Object] [ dynamic: true ]
Promise {
  <pending>,
  [Symbol(async_id_symbol)]: 5789,
  [Symbol(trigger_async_id_symbol)]: 5784,
  [Symbol(destroyed)]: { destroyed: false }
}
> Uncaught Error: syntax error at or near "["
```

### LINKED ISSUE #250: TypeError: Cannot read property 'length' of undefined    at Object.DataRow
A weird error happened on our server that seems to reveal a bug in the library.

Is a quite uncommon error that suggests is not an implementation problem but more a library bug.

for some reason under a now unknown circumstance `columns` seems to be undefined here:

https://github.com/porsager/postgres/blob/4a6669c1b623bb89fcf5a9d13dd09aa69c679811/lib/backend.js#L113

I am afraid that the information given may not be sufficient, but it can be an already known issue or something that may add up to another user information.

```sh
   for (let i = 0; i < backend.query.statement.columns.length; i++) {
                                                      ^
TypeError: Cannot read property 'length' of undefined    
at Object.DataRow (/opt/xxx/project/src/node_modules/postgres/lib/backend.js:113:57)    
at Socket.data (/opt/xxx/project/src/node_modules/postgres/lib/connection.js:319:25)    
at Socket.emit (node:events:365:28)    
at addChunk (node:internal/streams/readable:314:12)    
at readableAddChunk (node:internal/streams/readable:289:9)    
at Socket.Readable.push (node:internal/streams/readable:228:10)    
at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
error Command failed with exit code 1.
```

thanks

### LINKED ISSUE #252: unhandledRejection: Promise {   <rejected> TypeError [ERR_MISSING_ARGS]: The "options" or "port" or "path" argument must be specified       at new NodeError (node:internal/errors:329:5)       at Socket.connect (node:net:932:11)       at Object.connect (node:net:197:17)       at Object.connect (C:\GitHub\rcsng\nextjs\node_modules\postgres\lib\connection.js:392:13) {     code: 'ERR_MISSING_ARGS'   } }
```js
unhandledRejection: Promise {
  <rejected> TypeError [ERR_MISSING_ARGS]: The "options" or "port" or "path" argument must be specified
      at new NodeError (node:internal/errors:329:5)
      at Socket.connect (node:net:932:11)
      at Object.connect (node:net:197:17)
      at Object.connect (C:\...\node_modules\postgres\lib\connection.js:390:13) {
    code: 'ERR_MISSING_ARGS'
  }
}
```

This error occurs intermittently.

I have an unusual config here: `{ ..., idle_timeout:1, max:5 }`

Error is occuring here: https://github.com/porsager/postgres/blob/4a6669c1b623bb89fcf5a9d13dd09aa69c679811/lib/connection.js#L390

When error occurs it is because `i == 1` but I only have one host and port so `net.connect` is called with options `undefined, undefined` instead of the port number.

I think what is happening is multiple connections are trying to be reopened simultaneously

### LINKED ISSUE #254: INSERT with reserved column names ("order") needs escaping/quotes
The following breaks because `order` is a reserved POSTGRES keyword.

```js
const some_table_data = [
  { order: 1, transition: 'up' },
  { order: 2, transition: 'down' }
]
const results = sql`INSERT INTO some_table ${sql(some_table_data, 'order', 'transition')}`
```

This problem is similar to https://github.com/porsager/postgres/issues/21 in that the column `order` needs quotes to work.

Writing this SQL raw, should look something like this:

```sql
INSERT INTO some_table ("order",transition) values ($1,$2),($3,$4);
```

(Notice the quotes around `"order"`.

Instead, this library generates the following:
```sql
INSERT INTO some_table (order,transition) values ($1,$2),($3,$4);
```

Which throws a SQL syntax error.

I can't figure out a workaround other than to use `sql.unsafe`, which I'm forced to use to move forward, but I don't want that permanently. Any ideas?

porsager:
This is a complete rewrite to better support all the features that I was trying to get into v2. There are a few breaking changes from v2 beta , which some (myself included) are using in production, so I'm skipping a stable v2 release and going to v3.

Here are some of the new things available..
- Converted to ESM with CJS support
- Deno support
- Dynamic query builder based on raw sql
- Realtime subscribe to db changes through logical replication
- Cursors as async iterators
- Multi-host support for High Availability setups
- Postgres input parameter types from `ParameterDescription`
- `.describe()` to only get query input types and column definitions
- Support for Large Objects
- max_lifetime for connections
- Cancellation of requests

### Breaking changes from v2 -> v3
- Cursors are always called with `Result` arrays (previously cursor 1 would return a row object, where > 1 would return an array of rows)
- `.writable()` and `.readable()` is now async (returns a Promise that resolves to the stream)
- Queries now returns a lazy promise instead of being executed immediately. This means the query won't be sent until awaited (.then, .catch, .finally is called) or until `.execute()` is manually called.
- `.stream()` is renamed to `.forEach`
- Returned results are now it's own `Result` class extending `Array` instead of an Array with extra properties (actually shouldn't be breaking unless you're doing something funny)
- Parameters are now cast using the types returned from Postgres ParameterDescription with a fallback to the previously inferred types
- Only tested with node v12 and up
- Implicit array value to multiple parameter expansion removed (use sql([...]) instead)

### Breaking changes from v1 -> v2 (v2 never moved on from beta)
- All identifiers from `sql()` in queries are now always quoted
- Undefined parameters are no longer allowed
- Numbers that cannot be safely cast to JS Number are returned as string. This happens for eg, `select count(*)` because `count()` returns a 64 bit integer (int8), so if you know your `count()` won't be too big for a js number just cast in your query to int4 like `select count(*)::int`

Fixes #12, Fixes #30, Fixes #63, Fixes #65, Fixes #67, Fixes #89, Fixes #156, Fixes #159, Fixes #179, Fixes #201, Fixes #221, Fixes #230, Fixes #234, Fixes #248, Fixes #250, Fixes #252, Fixes #254
Closes #98, Closes #101, Closes #231, Closes #233


porsager:
I think the only thing left now is some missing and better documentation for the new features. Would be really great if anyone wants to help with that, since I'm not gonna be able to find a lot of time the next couple of weeks..

dilan-dio4:
@porsager how does the new "dynamic query builder based on raw sql" work? 

I'm very excited about that feature. Thanks!

porsager:
It's basically the most simply building block I could come up with for composing static sql safely into queries. I'm also very excited to see what people will come up with using it.

The ``` sql`` ``` instance can be used as a value inside a query to compose sql queries using safe static strings. 

I've also chosen this way to stay as close as possible to actual SQL.

Here are some examples for what you can do:

```js
// Using sql functions dynamically
sql` 
  update x set updated_at = ${ now ? sql`now()` : someDate }
`

// Dynamic filters (eg where, and / or usage)
sql` 
  select 
    * 
  from x ${ id 
    ? sql`where user_id = ${ id }` 
    : sql`` 
  }
`

// Complete queries using with 
const other = sql`
  select 1 as a
`

sql`
  with xs as (
    ${ other }
  )
  select * from xs
`
```

Now this should hopefully highlight that you can replace these with your own helper function(s) which should give you all the options you'd like. Now even though it does prevent sql injection of any kind, it doesn't mean you can't accidentally expose ways to access you database that was not intended, if you make any conditional query building available to users.

I hope this gives enough info to go on :)

dilan-dio4:
Okay this is great. I'm going to explore more later today. 

One more thing. I was having trouble reverse engineering where in code this takes place. Could you point me to where a dynamic query is recognized and how it is parsed?

porsager:
Sure thing :) This recursive thing does the work  https://github.com/porsager/postgres/blob/ece705a0f47d4967d0d098fa15ddcfc58e31cadd/src/connection.js#L235

dilan-dio4:
Awesome @porsager makes sense now.

dilan-dio4:
Here's what a few things needed for updated documentation: dynamic queries **(!)**, `.describe`, and async cursors.

With that I think we can simplify the structure of the README. There's a lot of stuff at the top that would serve better later in the document as being for advanced usage. I think there's an opportunity to make it easier for people get started with a simplified structure.

I propose a structure like this:

* Getting started
  * Installation
  * Usage
* **Basic** connection options
* Queries (this header will be for the basics of using sql\`\`)
  * Select
  * Insert
    * Multiple inserts with array of objects
  * Update
  * Delete
* Dynamic Queries
  * Usage
      *  Select, Insert, Update, Delete
  * Identifier names (sql(name))
  * Helpers (.json, arrays)
* Advanced iteration methods
  * .forEach
  * .cursor
  * .describe
  * .raw
  * .unsafe **(?)**
* Advanced communication
  * Listen and notify
  * .file
  * .subscribe
  *  Transactions
    * .begin
    * .savepoint
* Custom types
* postgres connection options
  * .postgres({ options }) **full** options object
  * SSL
  * Multi host connection - HA
  * Auto fetching of array types
  * Environment Variables for Options
  * The connection pool
* Error handling 
* TypeScript support

What's your thoughts on something like this @porsager ?

porsager:
Wow! That looks like a great start @dilan-dio4 , and I think the order and grouping you've come up with looks very nice!

Some notes:

- Perhaps "Advanced iteration methods" should be "Advanced query methods"
- sql.json helper is almost useless in v3 since you can pass json directly
- I like the position of unsafe, and then keep the collapsed version
- .file and transactions should probably go under "Advanced query methods"

I'll try to write the updated section for the connection pool description.

Minigugus:
@xiaoyu-tamu `npm install git+https://github.com/porsager/postgres.git#rewrite` :+1:

xiaoyu-tamu:
The promise never resolve with the new cursor implementation when cursor size > row size.


```
// assuming table has only 1 row

for await (const rows of sql`SELECT * FROM user;`.cursor(100)) {}

```

porsager:
Great catch @xiaoyu-tamu , I was missing tests for the new async iterator cursor mode, so this wasn't caught. I'll take a look at it.

jadbox:
Does v3 have TypeScript support for generating/reading query response types?

porsager:
Check out ``` sql``.describe() ``` 😊

karlhorky:
@porsager I'm guessing it's still a bit too early for our team to put v3 into our production workflow? Or do you think it's stable enough?

karlhorky:
Ah, tried it out, and after a few hours, an `ECONNRESET` error came up...

Wonder if it is related to the previous one: #180

```
Mar 7 08:10:45 PM  
Mar 7 08:10:45 PM  node:internal/process/promises:246
Mar 7 08:10:45 PM            triggerUncaughtException(err, true /* fromPromise */);
Mar 7 08:10:45 PM            ^
Mar 7 08:10:45 PM  Error: read ECONNRESET    at TCP.onStreamRead (node:internal/stream_base_commons:211:20)    at cachedError (file:///opt/render/project/src/node_modules/postgres/src/query.js:146:23)    at new Query (file:///opt/render/project/src/node_modules/postgres/src/query.js:32:55)    at sql (file:///opt/render/project/src/node_modules/postgres/src/index.js:98:11)    at getStudentByValidSessionToken (file:///opt/render/project/src/packages/pkg/build/database/queries/students.js:39:33)
Mar 7 08:10:45 PM  error Command failed with exit code 1.
Mar 7 08:10:45 PM  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Mar 7 08:10:46 PM  error Command failed.
Mar 7 08:10:46 PM  Exit code: 1
Mar 7 08:10:46 PM  Command: /opt/render/project/nodes/node-16.2.0/bin/node
Mar 7 08:10:46 PM  Arguments: /opt/yarn-v1.22.5/lib/cli.js start
Mar 7 08:10:46 PM  Directory: /opt/render/project/src/packages/pkg
Mar 7 08:10:46 PM  Output:
Mar 7 08:10:46 PM  
Mar 7 08:10:46 PM  info Visit https://yarnpkg.com/en/docs/cli/workspace for documentation about this command.
Mar 7 08:10:46 PM  error Command failed with exit code 1.
Mar 7 08:10:46 PM  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

porsager:
Huh. Well It also seems there's an issue with the error thrown, as it only shows the cached origin error 🧐

porsager:
Could you try adding `debug: true` too?

porsager:
What options do you have now?

karlhorky:
> What options do you have now?

Just `max`, set to ~20 average per service. I have 3 services, and Render allows 90 database connections.

> Could you try adding `debug: true` too?

Sure, will add that as well.

karlhorky:
> I'm going to try to add back the `idle_timeout: 100` configuration

Update: Going on 19 hours without incidents now.

So maybe the fix from `2.0.0-beta.6` (https://github.com/porsager/postgres/commit/60c1f8e7414f458cb9199e57c6ad2c920fccb390 ?) is no longer in v3?

porsager:
@karlhorky it's still there https://github.com/porsager/postgres/blob/rewrite/src/connection.js#L130 , so I don't think it's related.

Also v3 has a default max connection lifetime of 45-90 minutes .

But if you were running with 100 second idle_timeout I think you could quickly run into scenarios where you reconnect too often and at some point render will not allow a connection (hence the ECONNRESET).

karlhorky:
> But if you were running with 100 second idle_timeout I think you could quickly run into scenarios where you reconnect too often and at some point render will not allow a connection (hence the ECONNRESET).

I didn't have `idle_timeout: 100` before (no `idle_timeout` configured) - this is where it crashed twice in 4 hours.

I added `idle_timeout: 100` and since then, it hasn't crashed once in 19 hours.

karlhorky:
I'll reach out on Twitter to avoid spamming the thread too much

porsager:
How about https://gitter.im/porsager/postgres ?

porsager:
Hi @karlhorky 

I found out that setKeepAlive only works for the first connect. If the socket is reconnected there is no keep alive anymore. I'd say that it's a node bug, but it was a simple fix to move the `setKeepAlive` call to the connected handler. Would you mind trying again? :)

karlhorky:
Yep, trying out [the new version](https://github.com/porsager/postgres/pull/259/commits/5d90a32b8426af4802bc0fe324ddaa1d7964589d) now!

mohe2015:
Hey, @porsager. This is pretty impressive stuff and I love the ability to nest template strings arbitrarily. I have a use case though where I think I would need nested arrays of template strings. See https://github.com/projektwahl/projektwahl-lit/blob/9eb696c071eb111cc924253a5cc704faf470e4fc/src/server/entities.ts#L100 where I want to choose `ORDER BY` at runtime. My [current solution](https://github.com/projektwahl/projektwahl-lit/blob/9eb696c071eb111cc924253a5cc704faf470e4fc/src/server/sql/index.ts#L47) is a wrapper around v2 that is able to handle that (it looks horrifying mostly because of the typescript types).

Now that you added the ability to nest template strings I would be interested if you would like to add a solution to this here or if there already is one. I think there are mostly two possibilities:

1. Add a function to concatenate two tagged template strings. This should probably be comparably simple. Just as I'm writing this I'm realizing this is already possible. ```.reduce((prev, curr) => sql`${prev}${curr}`)``` obviously works.
2. Allow an array of these tagged template strings in a template string. This is probably comparably complicated and inefficient.

As I just realized this is already possible this is mostly a documentation request I think. I think the current simplicity is just better then and people could easily write a wrapper for that.

porsager:
@mohe2015 I think it'd be great with some more examples for things like that, but perhaps it's about time to have either an examples site / folder to show off idiomatic Postgres.js usage.

Currently I'm struggling just finishing the docs so v3 can be launched, so I'm not sure I'll have time for it, but any contributions are most welcome.

mohe2015:
> @mohe2015 I think it'd be great with some more examples for things like that, but perhaps it's about time to have either an examples site / folder to show off idiomatic Postgres.js usage.
> 
> Currently I'm struggling just finishing the docs so v3 can be launched, so I'm not sure I'll have time for it, but any contributions are most welcome.

I think I'll be able to create a PR to document such advanced usage. I think for the start we don't need too many examples with concatenation lots of cases are probably covered indirectly.

700software:
> Currently I'm struggling just finishing the docs so v3 can be launched, so I'm not sure I'll have time for it, but any contributions are most welcome.

@porsager, Does everything seem (even mostly) production-ready besides docs? We want to upgrade sooner than later.
(Having some weird connection closures even after [downgrading](https://github.com/porsager/postgres/issues/252#issuecomment-995226422).)

oskarrough:
For what it's worth we're also using v3 in production and so far so good. Thanks!

AVVS:
When fetching types fails an unhandled rejection is triggered:
(granted the query was waiting for 120s, error was captured for the outer query, but seems like there is a dangling promise somewhere?)

```
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^
PostgresError: query_wait_timeout
    at ErrorResponse (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:753:26)
    at handle (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:464:6)
    at TLSSocket.data (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:319:9)
    at TLSSocket.emit (node:events:390:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at TLSSocket.Readable.push (node:internal/streams/readable:228:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:199:23)
    at new Query (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/query.js:35:9)
    at fetchArrayTypes (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:710:25)
    at ReadyForQuery (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:523:16)
    at handle (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:464:6)
    at TLSSocket.data (/src/node_modules/.pnpm/github.com+porsager+postgres@3a8efe728057ca4c19437795af79bdbfd60baf4f/node_modules/postgres/cjs/src/connection.js:319:9)
    at TLSSocket.emit (node:events:390:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at TLSSocket.Readable.push (node:internal/streams/readable:228:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:199:23)
```

porsager:
I feel the docs are ready for release now. If anyone is up for it, I would love some feedback or just if you could read through them at https://github.com/porsager/postgres/tree/rewrite and check for any errors or missing details.

porsager:
Also just published an RC version to npm as well https://www.npmjs.com/package/postgres/v/3.0.0-rc.1

s0xDk:
**bytea** literals like `'\\xDEADBEEF'` are not working anymore.
Switched to `Buffer.from('DEAFBEEF', 'hex')`, but it's not quite succinctly.
Any other options? 🤔

porsager:
@s0xDk Hey, thanks a lot! 

Compared to v1?
Would you mind posting a quick sample how you used that before?

s0xDk:
@porsager 

having, for example, hex-encoded binary data:

```
const data = 'BAD455'
```

In **v1** and **v2.0.0-beta.11** I used:

```
await sql`INSERT INTO t SET b = ${`\\x${data}`}`
```

In **v3** that doesn't work. It's treating my already hex-encoded data as plain string `\xBAD455`, so I have to use:

```
await sql`INSERT INTO t SET b = ${Buffer.from(data, 'hex')}`
```

s0xDk:
Sometimes the following error throws up:

```
file://.../node_modules/postgres/src/connection.js:148
      return q.reject(Errors.connection('CONNECTION_DESTROYED', options))
                             ^

Error: write CONNECTION_DESTROYED /var/run/postgresql/.s.PGSQL.5432
    at Object.execute (file://.../node_modules/postgres/src/connection.js:148:30)
    at go (file://.../node_modules/postgres/src/index.js:348:14)
    at Promise.handler (file://.../node_modules/postgres/src/index.js:336:14)
    at Promise.handle (file://.../node_modules/postgres/src/query.js:128:65)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'CONNECTION_DESTROYED',
  errno: 'CONNECTION_DESTROYED',
  address: '/var/run/postgresql/.s.PGSQL.5432'
}

Node.js v17.7.2
```

But I don't use query timeouts, nor terminate any connections manually.

porsager:
> In **v1** and **v2.0.0-beta.11** I used:
> 
> ```
> await sql`INSERT INTO t SET b = ${`\\x${data}`}` // or
> await sql`INSERT INTO t SET b = ${'\\x'+data}`
> ```

Right, that's because v1 and v2 didn't know `b` was of type bytea, so the default serialisation was used (string). In v3 types are used so the serialization defined as [`serialize: x => '\\x' + Buffer.from(x).toString('hex')`](https://github.com/porsager/postgres/blob/rewrite/src/types.js#L37) for `bytea` will not work this way.

I unfortunately can't think of a nice way we can make both ways work. It's too lose to check if it's a string and if it starts with `\\x`, and it's also a bit weird if we expect strings to be hex.. I'm open to ideas...

porsager:
> ```
> Error: write CONNECTION_DESTROYED /var/run/postgresql/.s.PGSQL.5432
> ```

That is probably related to the new max lifetime, and the fact that a terminated connection is not removed immediately. Am I correct if this first comes after the service has run for 45 minutes?

Would you mind trying the latest in #rewrite ?

s0xDk:
> I unfortunately can't think of a nice way we can make both ways work. It's too lose to check if it's a string and if it starts with `\\x`, and it's also a bit weird if we expect strings to be hex.. I'm open to ideas...

something like this, maybe?
```js
serialize: x => x.startsWith('\\x') ? x : Buffer.from(x).toString('hex')
```

Or even more strict approach:
```js
serialize: x => x.match(/^\\x[0-9a-f]+$/i) ? x : Buffer.from(x).toString('hex') // or
serialize: x => x.match(/^\\x\p{Hex_Digit}+$/u) ? x : Buffer.from(x).toString('hex')
```

s0xDk:
> That is probably related to the new max lifetime, and the fact that a terminated connection is not removed immediately. Am I correct if this first comes after the service has run for 45 minutes?

Yes, it is. Sometimes it goes through that point without any issues, sometimes not.

Will try the #latest, thank you!

porsager:
> ```js
> serialize: x => x.startsWith('\\x') ? x : Buffer.from(x).toString('hex')
> ```
> 
> Or even more strict approach:
> 
> ```js
> serialize: x => x.match(/^\\x[0-9a-f]+$/i) ? x : Buffer.from(x).toString('hex') // or
> serialize: x => x.match(/^\\x\p{Hex_Digit}+$/u) ? x : Buffer.from(x).toString('hex')
> ```

Yeah, that's what I think is too loose. There's a really good chance no one ever will be bitten by it, but I wouldn't want to be responsible for it if so..

porsager:
> Will try the #latest, thank you!

Cool.. Let me know if it's good / bad.

s0xDk:
> Cool.. Let me know if it's good / bad.

Something new happened:

```
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^
Error: connect EISCONN /var/run/postgresql/.s.PGSQL.5432
    at PipeConnectWrap.afterConnect [as oncomplete] (node:net:1187:16)
    at cachedError (file://.../node_modules/postgres/src/query.js:158:23)
    at new Query (file://.../node_modules/postgres/src/query.js:36:9)
    at sql (file://.../node_modules/postgres/src/index.js:98:11)
    at my_func (file://.../my_script.js:392:25)

Node.js v17.7.2
```

karlhorky:
Thanks for merging and shipping @porsager, congratulations!! 🎉 🎊

porsager:
Thanks @karlhorky ! :) I hope it'll be put to good use ;)

arxpoetica:
AMAZING!!! YOU PUBLISHED v3!!! Can't wait to try it!

PR #264, opened 2022-02-06
Updated README

dilan-dio4:


dilan-dio4:
@porsager Here's where I'm at with restructuring some of the README. Let me know your thoughts.

porsager:
👏 I love it 😊 

cursor now also returns an async iterable if called without a callback function. 

I'll comment on the rest in line.

README.md
```diff
@@ -1,132 +1,109 @@
...
 - 🚯 1850 LOC - 0 dependencies
 - 🏷 ES6 Tagged Template Strings at the core
 - 🏄‍♀️ Simple surface API
+- 🖊️ Dynamic query support
 - 💬 Chat on [Gitter](https://gitter.im/porsager/postgres)
 
 <br>
 
+# Table of Contents
+
```
porsager, line 13: Should we perhaps move the table of contents below getting started?

README.md
```diff
@@ -145,378 +122,349 @@ const [new_user] = await sql`
...
-sql`
-  update users set ${
-    sql(user, 'name')
-  } where 
-    id = ${ user.id }
+const users = await sql`
+  select
+    *
+  from users
+  where age in (${ [68, 75, 23] })
```
porsager, line 345: This is deprecated in v3. It's done like this now by removing parenthesis and using sql():
```
in ${ sql(array) }
```

porsager (commented):
Ok, that's as far as i got on mobile, I have some more input, but will have to wait until I'm at my computer. 😊

README.md
dilan-dio4: Yes, will do

README.md
```diff
@@ -389,30 +389,23 @@ await sql`
 Use cursors if you need to throttle the amount of rows being returned from a query. New results won't be requested until the promise / async callback function has resolved.
 
 ```js
-
-await sql`
-  select * from generate_series(1,4) as x
-`.cursor(async ([row]) => {
+for await (const row of sql`select * from generate_series(1,4) as x`.cursor()) {
```
porsager, line 390: It's also always a result array being returned, even if a single row, so it needs to be `const [row] of ...`

porsager:
Ok, going a bit more thoroughly through it here is a list of some more things missing

- [ ] ``` const stream = await sql`copy x from stdin`.writable() ``` and ``` const stream = await sql`copy x to stdout`.readable() ```  is missing.

- [ ] Would be great with a description of the default `max_lifetime` setting which is a random interval for each connect between 45 and 90 minutes. This seemed to me like the best tradeoff between gaining the benefits of prepared statements, but still staying good friends with the OOM killer on linux. The rand9om interval was set to avoid connections reconnecting in bulk at the exact same time.

- [ ] Large objects usage documentation is not a must for v3 release, but would be nice.

- [ ] A description of the fact that queries are not executed until awaited or `.execute()` is called. This is necessary to support the usage of ``` sql`` ``` for fragments

- [ ] A mention of Deno support perhaps

README.md
dilan-dio4: Shoot I missed that lol

porsager: Hehe.. that's on me for just making these things with no traces ;)

dilan-dio4:
Changes made. I'll leave `stream` and Deno to you since I'm not too familiar in those spaces.

porsager:
Perfect, I think the rest should be manageable for me one of the coming days. 

Do you have more changes, or is merging fine now?

dilan-dio4:
Merging is good for now. If I have some more ideas over the next few days I'll add them as well.

porsager:
Awesome! Thanks again :)

dilan-dio4:
Of course! Do we have an ETA for v3 release?

porsager:
I don't have any outstanding issues now, so once I've added the rest to the docs we're good to go. During the coming week seems possible.

dilan-dio4:
Awesome!

PR #293, opened 2022-03-26
Update TypeScript types with v3 changes

eugene1g:
I'm sure this doesn't cover all changes from V3, but resolves some of the first issues I encountered.


porsager:
Ah - thanks a lot! Looks good

PR #294, opened 2022-03-27
Fix sql function overload type inference

### LINKED ISSUE #283: TypeScript error on await-ing sql`[...]`
```js
const sql = postgres(`postgres://${PGUSER}:${PGPASS}@${PGDOMAIN}:5432/tpp`, { debug });
const [u] = await sql`select id from users where twitterid = ${id}`;
```

Error:
```Type of 'await' operand must either be a valid promise or must not contain a callable 'then' member.```

Lib version: 3.0
Typescript: 4.6.3

This error did not occur in 2.0.

Minigugus:
Fixes #283

The SQL function is not simple to type because of all formats available. I tried a new approach here that seems to be more flexible and that give better error messages. However, all cases cannot be covered, especially with wrong or weak typing.

@jadbox Do you see some improvements?

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -151,6 +151,44 @@ type UnwrapPromiseArray<T> = T extends any[] ? {
   [k in keyof T]: T[k] extends Promise<infer R> ? R : T[k]
 } : T;
 
+type Keys = string
+
+type SerializableObject<T, K extends any[]> =
+  number extends K['length'] ? {} :
+  Record<Keys & (keyof T) & (K['length'] extends 0 ? string : K[number]), postgres.SerializableParameter>
+
+type First<T, K extends any[]> =
+  // Tagged template string call
+  T extends TemplateStringsArray ? TemplateStringsArray :
+  // Identifiers helper
+  T extends string ? string :
+  // Dynamic values helper (depth 2)
+  T extends readonly any[][] ? postgres.EscapableArray[] :
+  // Insert/update helper (depth 2)
+  T extends (object & infer R)[] ? SerializableObject<R, K>[] :
+  // Dynamic values helper (depth 1)
+  T extends readonly any[] ? postgres.EscapableArray :
+  // Insert/update helper (depth 1)
+  T extends object ? SerializableObject<T, K> :
+  // Unexpected type
+  never
+
+type Rest<T> =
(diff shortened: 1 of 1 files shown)


porsager:
How would types changes like this work with semver? Now that v3 is out, I'd like to make more frequent releases, but it'd be sad if having to do a major because of a Typescript change?

Minigugus:
> How would types changes like this work with semver? Now that v3 is out, I'd like to make more frequent releases, but it'd be sad if having to do a major because of a Typescript change?

I guess the best would be to updates types along with JS features so that breaking changes are merge into one.

> If it doesn't break anything but just fixes more edge cases it's fine in a patch right?

It adds smarter type inference (for instance, previously, `sql(user as any)` could have returned a query instead of an helper because TypeScript can't tell whether `user` is a string, an object or a `TemplateStringsArray`), adds the missing dynamic value case (`sql([['1', '2'], ['3', '4']])` (btw sorry I missed that new feature)) and allows nested queries (my bad too, sorry). I did my best to keep compatibility (```sql<ReturnType>`...`;``` should still work as expected), but I'd like some feedback to make sure types really work this time.

By the way, is `sql([[1, 2], [3, '4']])` valid? Since each nested array seams to correspond to a row (see below), then all nested arrays should have the same shape, and I could try to update types to detect that kind of issue where possible

https://github.com/porsager/postgres/blob/f7c8ae6ac63682ce747c0f645cde0c028cd020a6/tests/index.js#L973

porsager:
> I guess the best would be to updates types along with JS features so that breaking changes are merge into one.

For breaking changes sure, but for added features I don't want to tie updates down until types has been added (because I'm not the one doing them).

>  (btw sorry I missed that new feature))

Don't be sorry, I think there's a lot of typescript users out there who are very happy for your contributions and untangling my additions to v3 😉

If it's simply improvements and support for more cases, that's great - you should ping some of the users in the TS issues to try out your branch for feedback.

> By the way, is sql([[1, 2], [3, '4']]) valid?
Indeed it is, and yes they'd need to be the same shape.

porsager:
Would you say this is good to be merged @Minigugus ?

porsager:
I'll merge this @Minigugus - let me know if more needs changing ;)

porsager:
I'll merge this @Minigugus - let me know if more needs changing ;)

PR #300, opened 2022-03-31
Fix unsubscribing 

haikyuu:
Previously, unsubscribing was a no op because event listeneres were deleted from the wrong set.

Here is the script + command I was using for testing: `ps -eaf | grep walsender`
```JavaScript

const uri = 'env.PG_URI';
console.log(uri);
async function start() {

	const sql = postgres(uri, {
		// publications: 'custom_pub',
		fetch_types: false
	});
	const { unsubscribe: unsubscribe } = await sql.subscribe('*', function (row, { command: command, relation: relation, key: key, old: old }) {

		return console.log('event 1');
	});
	console.log("subscribed");
	return process.on('SIGINT', function () {

		console.log('unsub');
		return unsubscribe();
	});
};

start();

```

### DIFF
--- a/src/subscribe.js
+++ b/src/subscribe.js
@@ -31,11 +31,11 @@ export default function Subscribe(postgres, options) {
 
     const fns = listeners.has(event)
       ? listeners.get(event).add(fn)
-      : listeners.set(event, new Set([fn]))
+      : listeners.set(event, new Set([fn])).get(event)
 
     const unsubscribe = () => {
-      fns.get(event).delete(fn)
-      fns.get(event).size === 0 && listeners.delete(event)
+      fns.delete(fn)
+      fns.size === 0 && listeners.delete(event)
     }
 
     return connection.then(x => (stream = x, { unsubscribe }))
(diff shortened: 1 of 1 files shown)


porsager:
Ouch! Very nice catch - thank you.. It only happens on the first subscribe though so we need to change the fix a bit.

src/subscribe.js
```diff
@@ -34,8 +34,8 @@ export default function Subscribe(postgres, options) {
       : listeners.set(event, new Set([fn]))
```
porsager, line 34: Let's just add `.get(event)` at the end here as the fix instead.

src/subscribe.js
```diff
@@ -34,8 +34,8 @@ export default function Subscribe(postgres, options) {
...
 
     const unsubscribe = () => {
-      fns.delete(fn)
-      fns.size === 0 && listeners.delete(event)
+      fns.get(event).delete(fn)
+      fns.get(event).size === 0 && listeners.delete(event)
```
porsager, line 38: Remove `.get(event)` for these 2 lines and use above fix. With this fix it would break for additional subscribers added to the same event.

porsager (changes requested):

porsager (approved):

porsager:
Thanks a lot @haikyuu

PR #307, opened 2022-04-08
add code example to simplify conn. explanation

r-priyam:
Added code example in connection pool section to explain how postgres.js attempts to make the connection and reuse it, if needed.


porsager:
Awesome @r-priyam - thank you!

PR #309, opened 2022-04-08
document built in transform

r-priyam:
Document the best built-in method ✌️


r-priyam:
Tried my best to document it. Any suggestions are welcome 😃

PR #323, opened 2022-04-17
Remove unclear todo

alextes:
I'm guessing this was there when the dynamic query section was not yet and it is unneeded now, or maybe there was still something you wanted to add.

Either way, I'd suggest writing a full sentence on what is missing so that others can contribute or understand what information is missing or if nothing is needed just removing it. Right now it just feels confusing to the reader 😅 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ const xs = await sql`
 
 ### Query parameters
 
-Parameters are automatically extracted and handled by the database so that SQL injection isn't possible. No special handling is necessary, simply use tagged template literals as usual. **Dynamic queries and query building can be seen in the [next section]()**. // todo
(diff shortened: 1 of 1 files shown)


porsager:
Ah good catch. Yes, the entire text in bold can be removed since it's not necessary after restructuring the docs.

README.md
```diff
@@ -130,7 +130,7 @@ const xs = await sql`
 
 ### Query parameters
 
-Parameters are automatically extracted and handled by the database so that SQL injection isn't possible. No special handling is necessary, simply use tagged template literals as usual. **Dynamic queries and query building can be seen in the [next section]()**. // todo
+Parameters are automatically extracted and handled by the database so that SQL injection isn't possible. No special handling is necessary, simply use tagged template literals as usual. **Dynamic queries and query building can be seen in the [next section]()**.
```
porsager, line 133: We can remove all text in bold too.

alextes: Done 👍

porsager:
Nice - thanks

PR #332, opened 2022-04-25
add JAForbes/pgmg to the migration tools list

paulovieira:


porsager:
Cool - I'd love to merge that, but knowing @JAForbes let's just get his opinion if it's ready to include ☺️ (the version is < 1)

Is it fine for you @JAForbes ?

JAForbes:
Hey, yep happy for others to use it as long as they are ok with using a < 1.0 migration tool, but given ley and pgshift are both pre v1 I think it is fine in that context.

JAForbes:
Hey, yep happy for others to use it as long as they are ok with using a < 1.0 migration tool, but given ley and pgshift are both pre v1 I think it is fine in that context.

PR #338, opened 2022-04-28
Fix/improve sql() helper types

### LINKED ISSUE #331: Inserting object with no columns: No overload  matches this call
I just upgraded to 3.1.0 and this broke:
```js
await sql`INSERT INTO mytable ${ sql(object) }`;
```
Does not work. I have to list all the columns:
```js
await sql`INSERT INTO mytable ${ sql(object, 'col1', 'col2') }`;
```
Works as expected, but is much less convenient.

### LINKED ISSUE #335: cannot interpolate `Buffer[]` variable into query (TypeScript)
Using TypeScript, I'm trying to execute a query that updates rows while filtering via a `bytea` column, like so:
```typescript
// `txIdBuffers` is a const of type `Buffer[]`
await sql`
  UPDATE some_table
  SET val = true
  WHERE tx_id = ANY(${sql(txIdBuffers)})
`;
```
However, I'm getting the following error:
```
    Argument of type 'Buffer[]' is not assignable to parameter of type 'TemplateStringsArray'.
      Property 'raw' is missing in type 'Buffer[]' but required in type 'TemplateStringsArray'.
```
Could it be a matter of adding support to `Buffer[]` in `TemplateStringsArray`?
cc @Minigugus

Minigugus:
Fix #331, fix #335

- [X] `sql()`: allow `obj` to have properties not used in `props` and don't check types for these properties
- [X] `sql()`: fix the `Uint8Array` case (both a serializable value (`sql(buffer)`) and an object (`sql(buffer, 'length')`))
- [X] `sql()`: implicit JSON in objects
- [X] `sql.json()`: now check provided values (might be a breaking change in some cases, but would means there were already an issue with user types)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -112,7 +112,7 @@ interface BaseOptions<T extends JSToPostgresTypeMap> {
    */
   publications: string
   onclose: (connId: number) => void;
-  backoff: boolean | ((attemptNum:number) => number);
+  backoff: boolean | ((attemptNum: number) => number);
   max_lifetime: number | null;
   keep_alive: number | null;
 }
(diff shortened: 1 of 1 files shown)


porsager:
Does that also fix https://github.com/porsager/postgres/issues/335 ?

Minigugus:
Yes, sorry, I forgot that one :sweat_smile:

Maybe adding some kind of typescript label to relevant issues could help detect and fix them?

porsager:
Great idea 😊 I was just mentioning you instead 😋 

So, was that issue covered in this PR?

rafaelcr:
Thank you @porsager @Minigugus 🚀

PR #348, opened 2022-05-03
docs: add example of readable & writable COPY stream usage

zone117x:
Add documentation for the `<query>.writable()` and `<query>.readable()` APIs which are already covered in unit tests and typescript definitions, but not mentioned in the docs/readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,7 +451,7 @@ await pipeline(userStream, query);
 
 ```js
 const { pipeline } = require('stream/promises')
-const { createWriteStream } = require('fs');
+const { createWriteStream } = require('fs')
 
 const query = await sql`copy users (name, age) to stdin`.readable()
 await pipeline(query, createWriteStream('output.tsv'))
--- a/README.md
+++ b/README.md
@@ -428,9 +428,9 @@ const result = await sql.file('query.sql', ['Murray', 68])
 
 ### Rows as Streams
 
-Postgres.js supports [`copy ...`](https://www.postgresql.org/docs/14/sql-copy.html) queries, which are exposed as [Node.js streams](https://nodejs.org/dist/latest-v16.x/docs/api/stream.html).
+Postgres.js supports [`copy ...`](https://www.postgresql.org/docs/14/sql-copy.html) queries, which are exposed as [Node.js streams](https://nodejs.org/api/stream.html).
 
-> **NOTE** This is a low-level API which does not provide any type safety. To make this work, you must match your [`copy query` parameters](https://www.postgresql.org/docs/14/sql-copy.html) correctly to your [Node.js stream read or write](https://nodejs.org/dist/latest-v16.x/docs/api/stream.html) code. Ensure [Node.js stream backpressure](https://nodejs.org/en/docs/guides/backpressuring-in-streams/) is handled correctly to avoid memory exhaustion.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -426,6 +426,38 @@ Using a file for a query is also supported with optional parameters to use if th
...
+```
+
+#### ```await sql`copy ... to stdin` -> Readable```
+
+```js
+const { pipeline } = require('stream/promises')
+const { createWriteStream } = require('fs');
+
+const query = await sql`copy users (name, age) to stdin`.readable()
+await pipeline(query, createWriteStream('output.tsv'))
```
zone117x, line 457: This could alternatively be documented with an async iterator e.g. `for await (const user of query) { ... }`, but given the typical use cases for this, it seems like streams/pipes are a better example.

README.md
```diff
@@ -426,6 +426,38 @@ Using a file for a query is also supported with optional parameters to use if th
 const result = await sql.file('query.sql', ['Murray', 68])
 ```
 
+### Rows as Streams
+
+Postgres.js supports [`copy ...`](https://www.postgresql.org/docs/14/sql-copy.html) queries, which are exposed as [Node.js streams](https://nodejs.org/dist/latest-v16.x/docs/api/stream.html).
+
+> **NOTE** This is a low-level API which does not provide any type safety. To make this work, you must match your [`copy query` parameters](https://www.postgresql.org/docs/14/sql-copy.html) correctly to your [Node.js stream read or write](https://nodejs.org/dist/latest-v16.x/docs/api/stream.html) code. Ensure [Node.js stream backpressure](https://nodejs.org/en/docs/guides/backpressuring-in-streams/) is handled correctly to avoid memory exhaustion.
```
zone117x, line 433: It's tempting to mention the current default postgres COPY parameters here (i.e. tsv -- tabs for cells, new lines for rows), however, I think it's more useful to send readers to the postgres docs which cover this and more. 

Thoughts?

README.md
```diff
@@ -426,6 +426,38 @@ Using a file for a query is also supported with optional parameters to use if th
...
+const { pipeline } = require('stream/promises')
+
+// Stream of users with the default tab delimitated cells and new-line delimitated rows
+const userStream = Readable.from([
+  'Murray\t68\n',
+  'Walter\t80\n'
+])
+
+const query = await sql`copy users (name, age) from stdin`.writable()
+await pipeline(userStream, query);
```
zone117x, line 447: I'm following up the above docs about node.js stream backpressure with an example of using pipeline. It seems appropriate to me, but please let me know if this doc example should be changed to something else.

porsager:
Yes! Awesome - Thanks a lot.. Can't believe I missed that 🙈

README.md
porsager: Yeah, I agree with you - The PostgreSQL docs are great, but it can sometimes be hard to find what you're looking for if you aren't familiar with them. The link is great, but perhaps change `14` to `current` :)

README.md
porsager: Looks great 👍

README.md
porsager: Right, we could show both, it wouldn't take up much space..

zone117x: Added!

porsager:
Perfect - Thank you!

PR #354, opened 2022-05-08
docs: Convert CJS imports to ES

abdulrahman1s:
Converted cjs imports to es and added `node:` protocol to avoid confusion. [more about `node:`](https://github.com/nodejs/node/issues/38343)


porsager:
Thank you ;)

PR #360, opened 2022-05-12
Add optional `onlisten` callback to `listen()` on TypeScript

rafaelcr:
The TypeScript type definitions file currently lacks the optional `onlisten` callback for `sql.listen()`.
This PR adds the optional argument.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -608,7 +608,7 @@ declare namespace postgres {
     unsafe<T extends any[] = (Row & Iterable<Row>)[]>(query: string, parameters?: SerializableParameter[], queryOptions?: UnsafeQueryOptions): PendingQuery<AsRowList<T>>;
     end(options?: { timeout?: number }): Promise<void>;
 
-    listen(channel: string, cb: (value: string) => void): ListenRequest;
+    listen(channel: string, cb: (value: string) => void, onlisten?: () => void): ListenRequest;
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -608,7 +608,7 @@ declare namespace postgres {
     unsafe<T extends any[] = (Row & Iterable<Row>)[]>(query: string, parameters?: SerializableParameter[], queryOptions?: UnsafeQueryOptions): PendingQuery<AsRowList<T>>;
     end(options?: { timeout?: number }): Promise<void>;
 
-    listen(channel: string, cb: (value: string) => void): ListenRequest;
+    listen(channel: string, cb: (value: string) => void, onlisten?: () => void): ListenRequest;
```
Minigugus, line 611: Would it make sense to rename the first callback, for clarity?

```suggestion
    listen(channel: string, onnotify: (value: string) => void, onlisten?: () => void): ListenRequest;
```

Minigugus (commented):
Nice catch, thanks :+1:

types/index.d.ts
rafaelcr: Sounds good, applied suggestion 👍

porsager:
Thank you - I guess `onsubscribe` for subscribe is missing as well then?

rafaelcr:
@porsager you're right... sending a PR for that too

PR #361, opened 2022-05-12
Add implicit custom type inference

Minigugus:
Since `postgres` use oid for encoding parameters now, custom types no more need to be wrapped in `sql.types.someType(...)`, and then they should be infered with TypeScript (could help detect missing custom types by the way, for instance with `bigint`s)


porsager:
Thanks @Minigugus - I'll merge this to include in the next release.

PR #365, opened 2022-05-16
fix: deno assertRejects compatibility

baoshan:
`assertRejects` of Deno is defined as:

```
export function assertRejects<E extends Error = Error>(
  fn: () => Promise<unknown>,
  ErrorClass?: new (...args: any[]) => E,
  msgIncludes?: string,
  msg?: string,
): Promise<void>;
```

I suggest acknowledging `PostgresError` has a constructor by removing `private constructor()`. Please let me know if there are better approaches.

Thanks.

### DIFF
--- a/deno/types/index.d.ts
+++ b/deno/types/index.d.ts
@@ -227,9 +227,6 @@ declare namespace postgres {
     query: string;
     /** Only set when debug is enabled */
     parameters: any[];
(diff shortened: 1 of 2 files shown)


porsager:
Yes - Thank you! I don't see no reason why not to.

PR #370, opened 2022-05-18
Prevent error for undefined->null in array

e3dio:


e3dio:
This still does not work right, undefined gets inserted as "undefined" string, wanting to insert as Null, transform does not apply to array values ?

PR #373, opened 2022-05-20
fix: include missing boolean type in JSONValue union

thebearingedge:
This Pull Request adds `boolean` to the `JSONValue` type in `types/index.d.ts`.

image


porsager:
Looks good 😊  thanks

Issue #371, opened 2022-05-18
Inserting a Null (or Undefined with transform -> null) value in Array throws error


e3dio:
```javascript
await sql`CREATE TABLE test (x int[])`

const data = { x: [ 1, null, 3 ] }

await sql`INSERT INTO test ${sql(data)}`
```

```
TypeError: Cannot read properties of null (reading 'type')
    at \node_modules\postgres\src\types.js:247:49
```

e3dio:
Seems kind of a basic thing ability to insert a Null in array, library is throwing error, let me know if you want a PR

PR #375, opened 2022-05-24
Fix null in array

e3dio:
fix https://github.com/porsager/postgres/issues/371

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -243,7 +243,7 @@ export const arraySerializer = function arraySerializer(xs, serializer) {
   if (Array.isArray(first) && !first.type)
     return '{' + xs.map(x => arraySerializer(x, serializer)).join(',') + '}'
 
-  return '{' + xs.map(x => x === null ? 'null' :
+  return '{' + xs.map(x => x == null ? 'null' :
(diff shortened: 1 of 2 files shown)


src/types.js
```diff
@@ -243,7 +243,7 @@ export const arraySerializer = function arraySerializer(xs, serializer) {
   if (Array.isArray(first) && !first.type)
     return '{' + xs.map(x => arraySerializer(x, serializer)).join(',') + '}'
 
-  return '{' + xs.map(x =>
+  return '{' + xs.map(x => x == null ? 'null' :
```
zone117x, line 246: Doesn't this library only use strict equality comparisons?

porsager: Yes, this isn't going in.. I've already got a fix for this in the works, so closing the PR

porsager:
Something is already on the way, but thanks anyway ;)

src/types.js
e3dio: You use whatever equality comparison is needed, this fix works great

e3dio:
Ok could have said something like 5 days ago so I don't need to fix your library for you, thanks

PR #390, opened 2022-06-03
Running Unit Tests in Multiple Environments

Rlcolli4:
Unit tests are currently locked to a specific postgres set up that may already be in use in different development environments. The new template file and imports allow the tests to be ran in various environments where the dummy table may need to be ran on different ports or ip addresses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+tests/test_variables.js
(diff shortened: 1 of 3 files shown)


porsager:
Why can't you simply use environment variables as is? I don't think any of these changes are needed?

porsager:
Closing this for now, as I don't see why environment variables aren't enough to get the desired behavior.

Rlcolli4:
What if you have a database running on that port already for development purposes for a separate project? You don't think that you would want each person trying to contribute or work on this project to be able to have unique testing databases if needed?

Issue #386, opened 2022-06-01
Error: cannot cast type boolean to jsonb


uasan:
Hello.
Version `postgres 3.2.4`.

```JS
await sql`SELECT ${true}::jsonb`;
```

Result postgresql error:
```
Cannot cast type boolean to jsonb
```

Minigugus:
Maybe an issue with parameters types inference :thinking: On the meantime, maybe you can use `sql.json` instead?
```ts
await sql`SELECT ${sql.json(true)}::jsonb`;
```

Minigugus:
> I do not know the type of the parameter

Ok but your cast to `::jsonb` will make `postgres` call `JSON.stringify` anyway, so where is the problem? `sql.json` just returns an object that tells `postgres` which oid to use in the query, so if you always cast to JSONB it should work I guess

porsager:
@uasan oh shoot, that's actually not the only issue. Same goes for the other inferred types (Date, Buffer, Bigint) :-( I should have caught this for the v3 update and completely removed inferType (https://github.com/porsager/postgres/blob/218a7d4f37dcbf76d01081413a386c00544ab63e/src/types.js#L213) to let PostgreSQL decide :-/ Changing this would require a major bump, but my initial idea right now is that it would be the best direction.

uasan:
You, like many good developers, simply do not have enough tests for many cases, you need to think about how the community can help you write tests

uasan:
I think the best behavior in cases of `unknown` data type is to do a direct mapping of JS types to Postgres types, i.e.
```
JS boolean -> Postgres boolean
JS number -> Postgres numeric
JS bigInt -> Postgres numeric
```

AkeleyUA:
@porsager Hi. Should we expect a fix soon? If yes, how long?

P.S. I don't want to raise, but this is a pretty critical issue for us. Wish you happy coding.

porsager:
@AkeleyUA the issue is that we can't really change it without potentially breaking someone who relies on the implicit inference. It would require a new major, and I haven't thought through if that's the way to go yet. Could you maybe describe your specific issue / show an example query, and we can see if there's a way to work around it?

uasan:
> show an example query, and we can see if there's a way to work around it?

Dynamic query builder that knows the column name and values, but does not know the type of the column
```JS
sql` UPDATE SET "${sql(name)}" = ${value}`
```

> potentially breaking someone who relies on the implicit inference

[...]

PR #392, opened 2022-06-06
WIP: Remove js type inference

porsager:
In v3 we rely on the PostgreSQL ParameterDescription message to decide on input types, but there is still some cases were we rely on types inferred from the js types first (https://github.com/porsager/postgres/blob/218a7d4f37dcbf76d01081413a386c00544ab63e/src/types.js#L213-L223). This [causes problems](https://github.com/porsager/postgres/issues/386) when trying to use some of these JS types as raw values for eg. `json` PostgreSQL types (and others). Because of the order of the protocol messages we are not able to give "hints" about types so it could be better to remove the "convenience" of implicitly inferring types from the js objects and instead require the users to write explicit casts in sql - eg `::type or cast(...)`. 

I'm still not certain this is the best way forward, and if someone sees an alternative solution that encompasses both scenarios - i am all ears.

For now, this PR can act as a test ground and discussion for the changed behaviour.

### DIFF
--- a/cjs/src/types.js
+++ b/cjs/src/types.js
@@ -211,15 +211,11 @@ const escapeIdentifier = module.exports.escapeIdentifier = function escape(str) 
 }
 
 const inferType = module.exports.inferType = function inferType(x) {
-  return (
-    x instanceof Parameter ? x.type :
-    x instanceof Date ? 1184 :
-    x instanceof Uint8Array ? 17 :
-    (x === true || x === false) ? 16 :
-    typeof x === 'bigint' ? 20 :
-    Array.isArray(x) ? inferType(x[0]) :
-    0
-  )
+  return x instanceof Parameter
+    ? x.type
+    : Array.isArray(x)
+      ? inferType(x[0])
+      : 0
 }
 
 const escapeBackslash = /\\/g
--- a/cjs/tests/index.js
+++ b/cjs/tests/index.js
@@ -89,16 +89,16 @@ t('String', async() =>
 )
 
 t('Boolean false', async() =>
-  [false, (await sql`select ${ false } as x`)[0].x]
+  [false, (await sql`select ${ false }::bool as x`)[0].x]
 )
 
 t('Boolean true', async() =>
-  [true, (await sql`select ${ true } as x`)[0].x]
+  [true, (await sql`select ${ true }::bool as x`)[0].x]
 )
 
 t('Date', async() => {
   const now = new Date()
-  return [0, now - (await sql`select ${ now } as x`)[0].x]
+  return [0, now - (await sql`select ${ now }::timestamptz as x`)[0].x]
 })
 
 t('Json', async() => {
@@ -134,7 +134,8 @@ t('Array of String', async() =>
 
 t('Array of Date', async() => {
   const now = new Date()
-  return [now.getTime(), (await sql`select ${ sql.array([now, now, now]) } as x`)[0].x[2].getTime()]
+      , iso = now.toISOString()
+  return [now.getTime(), (await sql`select ${ sql.array([iso, iso, iso]) }::timestamptz[] as x`)[0].x[2].getTime()]
 })
 
 t('Nested array n2', async() =>
@@ -1034,10 +1035,7 @@ t('Multiple queries', async() => {
 })
 
 t('Multiple statements', async() =>
-  [2, await sql.unsafe(`
-    select 1 as x;
-    select 2 as a;
-  `).then(([, [x]]) => x.a)]
+  [2, await sql.unsafe(`select 1 as x;select 2 as a`).then(([, [x]]) => x.a)]
 )
 
 t('throws correct error when authentication fails', async() => {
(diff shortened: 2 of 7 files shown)


uasan:
> require the users to write explicit casts in sql - eg ::type or cast(...).

this is a very correct requirement, in the `TS` this requirement does not surprise anyone.

It is always easier when there is only one source of truth, postgres data types are needed by postgres, so at the postgres level (SQL) you need to describe types

uasan:
Here are some more examples of comparing the two ways of describing types and inferring types:

`bytea` from JS hex string
```JS
// describe type
sql`SELECT ${'\\x27'}::bytea`;

// inference type
sql`SELECT ${Buffer.from('\\x27', 'hex')}`;
```

`date` from JS string
```JS
// describe type
sql`SELECT ${'2020-10-11'}::date`;

// inference type
sql`SELECT ${new Date('2020-10-11')}`;
```

`bigint` from JS number
```JS
// describe type
sql`SELECT ${123}::bigint`;

// inference type
sql`SELECT ${BigInt(123)}`;
```

As you can see, if your values have core values for postgres types, but you have to explicitly convert these values into correspondence types of JS, it's easier to just specify these types in the SQL itself

porsager:
@uasan well that's what I've been trying to get at the whole time. We can only choose one over the other, but I feel like you keep suggesting that we can have both? Can you please confirm that you also don't see any way to have both ? ☺️

For instance if we switch to no inference we can't do this:
```js
sql`
  select * from x
  where ${ someBoolean } or user_id = ${ user_id }
`
```
But will have to do this:

```js
sql`
  select * from x
  where ${ someBoolean }::boolean or user_id = ${ user_id }
`
```

uasan:
> We can only choose one over the other, but I feel like you keep suggesting that we can have both? Can you please confirm that you also don't see any way to have both ?

My suggestion is we should first of all determine types from the description SQL query, but if type is unknown type (as in your second example), for backwards compatibility, we can determine the type on based on the JS data type, but this second source of truth is needed only for backwards compatibility and for unknown types from description SQL, i.e. in rare cases.

porsager:
But you'll never get the db to see that as anything but a unknown type. The issue is the order required of the extended query protocol.

uasan:
Probably because of my bad English, you did not fully understand my idea.
If the data type is not known, you need to pass the data as an unknown type, but serelize to text data according to the JS data type, i.e.

```
true -> 't'
date -> date.toDateString()
...
```
You do not need to send a request for parameter descriptions again, you need to send a BIND with an unknown type, but with correct text serialization, which is performed based on the JS of the input parameter value types

Minigugus:
What about a flag in options? I mean introducing this feature as an opt-in (e.g. a optional `inferTypesFrom?: 'postgres' | 'js'` property in options, which default to `'js'` in v3), then remove the flag in the next major (or set it to `'postgres'` by default)?

porsager:
@uasan - english isn't my first language either ;)

 I thought you meant something else since I think what you describe is what this PR does :P So sorry if I'm being obtuse, but just want to clarify that when we send correctly serialized value, but unknown type, postgres won't know about the type anyway. Specifically the following `true` values will be seen simply as a string.

```
sql`
  select * from x
  where ${ someBoolean } or user_id = ${ user_id }
`
```

@Minigugus yes - pragmatism for the win!! But yeah, now we have a new problem - the name for that option 😝 I'm not sure I think `inferTypesFrom` and an enum is necessary. How about `strict_types: true | false`?

uasan:
> Specifically the following true values will be seen simply as a string

I didn't check your example, but I think postgres will determine the type of `someBoolean` variable as boolean and give this `oid` in the response from describe

porsager:
Thanks a lot for sticking with me @uasan - You are right! In that example it'll know. So would you think it's only for the pass-through selects we'll loose the type? eg: `select ${ true }`

uasan:
Yes, unknown type, this is quite rare, most of these requests have a design-level error, it's especially fun when using such untyped parameters to call SQL functions with overloading on the types of arguments of this function )

uasan:
About `null` value json.
```JS
UPDATE table
SET json_column = ${null}::json
```
now all null values are sent as SQL null, we have only one option how to send `'null'` values as JSON values using helper `sql.json(null)`?

porsager:
@uasan that's a good point too - could be quite confusing since sql null and json null would translate to the same js null when read out again..

uasan:
> Got any ideas for solving it cleanly?

this is a value collision, SQL null is more like JS undefined, null is JSON values, but this breaks backwards compatibility a lot, I think developers should resolve their own value collisions

peterbe:
I'll be honest, I didn't understand the TypeScript error. In the end, I solved it with `// @ts-ignore` and it's working fine. 

```ts
        // @ts-ignore
        await sql`
          insert into base_requestlog
           (url, status_code, created, request, response, meta)
          values (
            ${data.request.url.slice(0, 500)}, ${data.response.statusCode}, NOW(),
            ${data.request}, 
            ${data.response}, 
            ${data.meta}
          )
        `
```
Here, this `data` object is a plain an simple object. Something like:
```ts
const data = {
  request: {url: "/foo"},
  response: {statusCode: 200},
  meta: {hello: "world"}
}
```

PR #398, opened 2022-06-12
Adjustment for better wording.

DavidPesta:


porsager:
Replacing it with intervention removes the intended meaning. 
I think simply ending the sentence after `seamlessly` is better then.

e3dio:
I think when it says "without user interference" or "without user intervention" it is meaning to say "automatically" which is a more accurate term to use, probably skipped it because it immediately follows "seamlessly" which would make it 2 "ly" adverbs in a row, somewhat awkward but still doable "come up and down seamlessly automatically"

DavidPesta:
I updated it to use both words "seamlessly" and "automatically" without both appearing next to one another.

e3dio:
Ah he already used "automatically" in the same sentence already taken nevermind

e3dio:
Ok your edit is good I think

e3dio:
Yeah too many automatically's in the same paragraph, not good. Probably intervention is more correct than interference

DavidPesta:
no problem, changing it back to what i had originally then

porsager:
> This allows multiple connections to seamlessly come up and down independently without affecting the service.

This is more what I had in mind.

DavidPesta:
Oh, I thought point being made was that it happens automatically without the user/developer needing to do anything to handle doing it.

e3dio:
It's already stated in section above that connections happen automatically so I think it's implied, but "without affecting the service" makes sense probably what was originally meant

DavidPesta:
Ok, I like the clarity that the new wording adds. The words "user interference" was bothering me. This is probably good now.

porsager:
👍

Issue #407, opened 2022-06-29
TypeError: doesn't work if `exactOptionalPropertyTypes:true` in tsconfig.json


xerosanyam:
`exactOptionalPropertyTypes` just makes type system a little stricter. It is a good to have flag, suggested by [zapatos](https://jawj.github.io/zapatos/).

The error that's currently thrown is:
<img width="1345" alt="image" src="https://user-images.githubusercontent.com/3167278/176410444-dcf19060-2b16-46f0-b724-b5a6de76a365.png">

porsager:
I'm not into Typescript, but perhaps @Minigugus could share a comment? :)

PR #416, opened 2022-07-03
Minor typescript organization

thebearingedge:
* toJSON() of JSONValue returns JSONValue type
* TTypes defaults to JSToPostgresTypMap
* JSToPostgresTypeMap is in postgres namespace

The JSONValue type already includes recursive type references, so we may as well
guarantee that JSON.stringify() will get a JSONValue back.

Having to always include {} when referencing Sql<{}> is annoying.

This exposes the type parameter of the Sql interface for extension by other libraries that would like to extend it.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,7 +5,7 @@ import { Readable, Writable } from 'node:stream'
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList = {}>(options?: postgres.Options<T>): postgres.Sql<{ [type in keyof T]: T[type] extends {
+declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(options?: postgres.Options<T>): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
   serialize: (value: infer R) => any,
   parse: (raw: any) => infer R
 } ? R : never }>
@@ -15,7 +15,7 @@ declare function postgres<T extends PostgresTypeList = {}>(options?: postgres.Op
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList = {}>(url: string, options?: postgres.Options<T>): postgres.Sql<{ [type in keyof T]: T[type] extends {
+declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(url: string, options?: postgres.Options<T>): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
   serialize: (value: infer R) => any,
   parse: (raw: any) => infer R
 } ? R : never }>
@@ -23,7 +23,7 @@ declare function postgres<T extends PostgresTypeList = {}>(url: string, options?
 /**
  * Connection options of Postgres.
  */
-interface BaseOptions<T extends PostgresTypeList> {
+interface BaseOptions<T extends Record<string, postgres.PostgresType>> {
   /** Postgres ip address[s] or domain name[s] */
   host: string | string[];
   /** Postgres server[s] port[s] */
@@ -126,7 +126,6 @@ interface BaseOptions<T extends PostgresTypeList> {
   keep_alive: number | null;
 }
 
-interface PostgresTypeList extends Record<string, postgres.PostgresType> {}
 
 declare const PRIVATE: unique symbol;
 
@@ -277,7 +276,7 @@ declare namespace postgres {
     [name: string]: string;
   }
 
-  interface Options<T extends PostgresTypeList> extends Partial<BaseOptions<T>> {
+  interface Options<T extends Record<string, postgres.PostgresType>> extends Partial<BaseOptions<T>> {
     /** @inheritdoc */
     host?: string;
     /** @inheritdoc */
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,7 +5,7 @@ import { Readable, Writable } from 'node:stream'
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList>(options?: postgres.Options<T>): postgres.Sql<PostgresTypeList extends T ? {} : { [type in keyof T]: T[type] extends {
+declare function postgres<T extends PostgresTypeList = {}>(options?: postgres.Options<T>): postgres.Sql<{ [type in keyof T]: T[type] extends {
   serialize: (value: infer R) => any,
   parse: (raw: any) => infer R
 } ? R : never }>
@@ -15,7 +15,7 @@ declare function postgres<T extends PostgresTypeList>(options?: postgres.Options
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList>(url: string, options?: postgres.Options<T>): postgres.Sql<PostgresTypeList extends T ? {} : { [type in keyof T]: T[type] extends {
+declare function postgres<T extends PostgresTypeList = {}>(url: string, options?: postgres.Options<T>): postgres.Sql<{ [type in keyof T]: T[type] extends {
   serialize: (value: infer R) => any,
   parse: (raw: any) => infer R
 } ? R : never }>
@@ -126,9 +126,7 @@ interface BaseOptions<T extends PostgresTypeList> {
   keep_alive: number | null;
 }
 
-interface PostgresTypeList {
-  [name: string]: postgres.PostgresType;
-}
+interface PostgresTypeList extends Record<string, postgres.PostgresType> {}
 
 declare const PRIVATE: unique symbol;
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -130,10 +130,6 @@ interface PostgresTypeList {
   [name: string]: postgres.PostgresType;
 }
 
-interface JSToPostgresTypeMap {
-  [name: string]: unknown;
-}
-
 declare const PRIVATE: unique symbol;
 
 declare class NotAPromise {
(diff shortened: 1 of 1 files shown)


porsager:
Would you mind taking a look @Minigugus ?

types/index.d.ts
```diff
@@ -480,7 +476,7 @@ declare namespace postgres {
     | boolean
     | Date // serialized as `string`
     | readonly JSONValue[]
-    | { toJSON(): any } // `toJSON` called by `JSON.stringify`; not typing the return type, typings is strict enough anyway
+    | { toJSON(): JSONValue } // `toJSON` called by `JSON.stringify`
```
Minigugus, line 479: Types are already quite strict in this library, and enforcing types for `toJSON()` may cause troubles when integrating with others libraries that don't respect the `JSONValue` type definition. Anyway, it's not `postgres` responsibility to validate user-side `toJSON` implementations: users that care about proper types should type the `toJSON` definition instead.

```suggestion
    | { toJSON(): any } // `toJSON` called by `JSON.stringify`; not typing the return type, types definition is strict enough anyway
```

Minigugus (changes requested):

types/index.d.ts
```diff
@@ -581,7 +577,11 @@ declare namespace postgres {
     rest: U;
   }
 
-  interface Sql<TTypes extends JSToPostgresTypeMap> {
+  interface JSToPostgresTypeMap {
+    [name: string]: unknown;
+  }
+
+  interface Sql<TTypes extends JSToPostgresTypeMap = JSToPostgresTypeMap> {
```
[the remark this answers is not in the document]
thebearingedge: No problem 👌

types/index.d.ts
thebearingedge: Ok, I think that's fair 👍

thebearingedge:
@Minigugus Thank you for the review!

Minigugus (changes requested):
~~I guess you've tried to rebase or something? :thinking:~~

Ok you just ran the `build` script, sorry :sweat_smile: Maybe it would be better to either commit the `build` script result before merging, or at least only push types changes?

thebearingedge:
@Minigugus Sorry, I believe that the build script ran due to the `prepare` script, but I am not accustomed to committing build artifacts so I didn't explicitly add it.

I'm pretty sure I botched synchronizing my fork with upstream, hence the rebase. I am going to attempt to clean things up this morning.

Minigugus:
@thebearingedge Do you think you could change `PostgresTypeList` to `Record<string, postgres.PostgresType>` too? It's quite similar to the fix for `JSToPostgresTypeMap` so it is relevant to fix `PostgresTypeList` here too? #448 would be solved by this issue then

RealShadowNova (approved):

thebearingedge:
@Minigugus yes, I should be able to take care of that 👍

types/index.d.ts
```diff
@@ -126,13 +126,7 @@ interface BaseOptions<T extends PostgresTypeList> {
...
 }
 
-interface PostgresTypeList {
-  [name: string]: postgres.PostgresType;
-}
-
-interface JSToPostgresTypeMap {
-  [name: string]: unknown;
-}
+interface PostgresTypeList extends Record<string, postgres.PostgresType> {}
```
thebearingedge, line 129: I have updated the `PostgresTypeList` to extend a `Record`.

types/index.d.ts
```diff
@@ -15,7 +15,7 @@ declare function postgres<T extends PostgresTypeList>(options?: postgres.Options
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList>(url: string, options?: postgres.Options<T>): postgres.Sql<PostgresTypeList extends T ? {} : { [type in keyof T]: T[type] extends {
+declare function postgres<T extends PostgresTypeList = {}>(url: string, options?: postgres.Options<T>): postgres.Sql<{ [type in keyof T]: T[type] extends {
```
thebearingedge, line 18: I am happy to change this, but I could not figure out why there was a condition for `PostgresTypeList extends T` and when I tried my fork it seemed to have the same effect when omitted.

thebearingedge (commented):
@Minigugus I made a modification to the signatures of `postgres<T>`, but am happy to undo it. There may be some reason for the condition that I am unaware of.

types/index.d.ts
Minigugus: The issue with `PostgresTypeList` is that it's confusing for many users, so the best would be to just replace this type with `Record<string, postgres.PostgresType>`, just like you did with `JSToPostgresTypeMap` :+1:

types/index.d.ts
Minigugus: This is because sometimes TS infer `PostgresTypeList` (and not `{}`) under some conditions, then the `keyof T` is replace with `string` in the following object literal, and the `TTypes` generic become just a `Record<string, CustomTypeA | CustomTypeB | ...>`, which is quite useless. But maybe recent versions of typescript have fixed that, I checked again recently

thebearingedge:
@Minigugus Ok! I think we are all set 🔥 ⌨️ 🔥

porsager:
Is this ready to merge once conflicts have been resolved @Minigugus ?

Minigugus (approved):
Sorry for the late reply

Looks good to me :+1:

porsager:
Would any of you mind resolving the current conflicts ? :)

thebearingedge:
Yes I will tackle it sometime this week 👍

thebearingedge:
@porsager thank you, the conflict has been resolved :salute:

types/index.d.ts
```diff
@@ -5,7 +5,7 @@ import { Readable, Writable } from 'node:stream'
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList>(options?: postgres.Options<T> | undefined): postgres.Sql<PostgresTypeList extends T ? {} : { [type in keyof T]: T[type] extends {
+declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(options?: postgres.Options<T>): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
```
Minigugus, line 8:
```suggestion
declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(options?: postgres.Options<T> | undefined): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
```

The `| undefined` part is on purpose: #407

types/index.d.ts
```diff
@@ -15,15 +15,15 @@ declare function postgres<T extends PostgresTypeList>(options?: postgres.Options
  * @param options Connection options - default to the same as psql
  * @returns An utility function to make queries to the server
  */
-declare function postgres<T extends PostgresTypeList>(url: string, options?: postgres.Options<T> | undefined): postgres.Sql<PostgresTypeList extends T ? {} : { [type in keyof T]: T[type] extends {
+declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(url: string, options?: postgres.Options<T>): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
```
Minigugus, line 18:
```suggestion
declare function postgres<T extends Record<string, postgres.PostgresType> = {}>(url: string, options?: postgres.Options<T> | undefined): postgres.Sql<Record<string, postgres.PostgresType> extends T ? {} : { [type in keyof T]: T[type] extends {
```

Same here

Minigugus (approved):
Just add theses missing `undefined` and it's ready to merge :slightly_smiling_face:

akatechis:
Would this PR fix the error `type "string" does not exist`? I'm using postgresjs with the following code example:

```typescript
import postgres from 'https://deno.land/x/postgresjs@v3.1.0/mod.js';

const sql = postgres<PGTypes>({
  host: 'localhost',
  port: 5432,
  database: 'postgres',
  user: 'postgres',
  password: 'postgres',
});

interface PGTypes {
  string: 'text';
  datetime: 'timestamp';
  integer: 'int';
  [name: string]: string;
}
```

But when I run it, I get the following error, and something tells me I'm not passing the correct `JSToPostgresTypeMap` type parameter to the `postgres` factory function...

```
error: Uncaught (in promise) PostgresError: type "string" does not exist
    const error = Errors.postgres(parseError(x))
                         ^
    at ErrorResponse (https://deno.land/x/postgresjs@v3.1.0/src/connection.js:761:26)
    at handle (https://deno.land/x/postgresjs@v3.1.0/src/connection.js:468:6)
    at data (https://deno.land/x/postgresjs@v3.1.0/src/connection.js:312:9)
    at https://deno.land/x/postgresjs@v3.1.0/polyfills.js:140:32
    at Array.forEach (<anonymous>)
    at call (https://deno.land/x/postgresjs@v3.1.0/polyfills.js:140:18)
    at success (https://deno.land/x/postgresjs@v3.1.0/polyfills.js:100:11)
```

If there's any documentation on what to pass here, I would gladly accept a link so I can read up on how to pass a correct generic interface

thebearingedge:
Hi @akatechis, The `types` option will need to be an object, like this example under Custom Types.

https://github.com/porsager/postgres#custom-types

akatechis:
Thanks, @thebearingedge, i'll look into that. Does that require me to give type definitions for basic types like string as well, or is that just for custom types? The table I'm trying to query is just two string columns, created by the following:

```typescript
sql`create table if not exists ${sql(tableName)} (id string not null primary key, data string)`
```

I would assume I don't need to define how to handle strings. Is that correct?

thebearingedge:
@akatechis no worries 😄 

regarding your custom `string` type. for that you will need to define it as a domain, as i do not believe that postgres has it. there is a `text` type though.

```sql
create domain string as text;
```

https://www.postgresql.org/docs/current/sql-createdomain.html

the custom type settings for the `postgres` package are there mainly to let you pre-process and post-process values transferring to/from your JavaScript code. so you might have something like this (just a random example)

```ts
const types = {
  string: {
    to: 25, // the PG oid for "text"
    from: [25],
    serialize: s => s.toLowerCase(), // lowercase before storing in PG
    parse: s => s.toUpperCase()  // uppercase before using in JS
  }
}
```

PR #422, opened 2022-07-05
Update README.md

xerosanyam:
`no_prepare` was deprecated.


porsager:
Ah - good catch.. Perhaps even change to 

> This can be disabled by setting `prepare: false`

?

xerosanyam:
thanks for suggestion! updated.

PR #436, opened 2022-07-22
fix #429 documentation of lazy promise

### LINKED ISSUE #429: document: queries not executed until you call .then
Hey I'm posting what I've gathered here so hopefully it will show up on [Google Search](https://www.google.com/search?q=porsager+postgres+doesn%27t+execute+until+you+call+then) because right now it's a bit hard to find something.

From Gitter conversation: (https://gitter.im/porsager/postgres?at=62d5b60f0a522647984b1591):

> Bryan Field @700software 2022-07-18
> I have some updates that run async in background without using await
> 
> ```ts
> // async
> sql`update auth_session set activity=now() where id=...`
> but nothing happens.
> ```
> 
> However, if I add `.then()` now it works.
> 
> Noting this is mentioned by @porsager Jun 21 above:
> 
> > Queries are not executed until either, `then`, `catch`, `finally` or `execute` is called. `await` implicitly calls then.
> 
> 1 . Is this functionality documented anywhere on the library docs?
> 2 . Why is it this way?
> 3 . Any way to switch this behavior off?
> It seems inconsistent to any other promises.
> For example `const promise = new Promise((resolve, reject) => console.log('foo'))` seems to invoke `console.log` immediately without waiting for `.then` to be called. Yours is the first library that seems to work around this behavior and it is a convention I've never seen.
> 
> Bryan Field @700software 15:49
> Seeing it documented here: https://github.com/porsager/postgres/blob/599d0e7/README.md?plain=1#L130
> 
> Rasmus Porsager @porsager 15:54
> This behavior is what allows using <code>sql``</code> as fragments ;)
> 
> Bryan Field @700software 16:01
> What!? I can do that?
> I will look into
> I so have been wanting fragments.
> So @porsager a little background we actually have a middleman function that blocks all SQL requests until certain tablechanges have intervened. :)
> I will think about updating the middleman to support this. Pretty cool.
> So the middleman is not using lazy promises yet but found when using lib directly e.g. no tablechanges needed then lazy promises are the norm.
> 
> Rasmus Porsager @porsager 16:05
> Yeah you can do eg:
> ```ts
> const xs = await sql`
>   select * from users
>   order by name ${ asc ? sql`asc` : sql`desc` }
> `
> ```
> And a lot more advanced stuff follows too :P
> ...
> 
> Bryan Field @700software 16:07
> Listening to this lazy promise logic described here: https://changelog.com/jsparty/221#t=22:08.
> 
> Rasmus Porsager @porsager 16:07
> Ah, I forgot we touched on that in the talk..

So we can leave it here as an issue or move it to more official docs. But either way, at least now Google Searches will more likely find this.

700software:
fix #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ async function insertUser({ name, age }) {
 
 * [Connection](#connection)
 * [Queries](#queries)
-* [Building queries](#building-queries)
+* [Building queries with nested Fragments](#building-queries-with-nested-fragments)
 * [Advanced query methods](#advanced-query-methods)
 * [Transactions](#transactions)
 * [Data Transformation](#data-transformation)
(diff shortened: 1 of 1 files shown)


porsager:
I don't want to encourage not handling errors in the docs, so no shoot and forget samples using execute() like that please 😊

porsager:
I

PR #439, opened 2022-07-25
Add typescript support for "exports"

tarantoj-intrepid:
Allows typescript to properly resolve the types for this package when using the new Node.js’ ECMAScript Module Support

https://www.typescriptlang.org/tsconfig#moduleResolution

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "module": "src/index.js",
   "main": "cjs/src/index.js",
   "exports": {
+    "types": "./types/index.d.ts",
     "import": "./src/index.js",
     "default": "./cjs/src/index.js"
   },


Minigugus (approved):

Snurppa:
Just adding for possible error search match, if people happen to end up here with `Could not find a declaration file for module 'postgres'` errors when using TypeScript `NodeNext` module resolution.

Error with current `3.2.4` version, without this fix merged in is:
```
api/src/components/db.ts:3:22 - error TS7016: Could not find a declaration file for module 'postgres'. '/Users/x/y/z/node_modules/postgres/src/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/postgres` if it exists or add a new declaration (.d.ts) file containing `declare module 'postgres';`

  import postgres from 'postgres';
                       ~~~~~~~~~~
```

I guess declaring `declare module 'postgres'` yourselves could be a solution while waiting for release, but I settled for brtual edit of `node_modules/postgres/package.json` and added related `exports.types` entry by hand.
Not production grade fix, but works for me as I am only developing locally currently.

PR #447, opened 2022-07-30
Fix some `undefined` types issues

### LINKED ISSUE #407: TypeError: doesn't work if `exactOptionalPropertyTypes:true` in tsconfig.json

Minigugus:
Fixes #407 `undefined` were missing on optional properties
Fixes #446 `undefined` were not allowed in insert helper calls

Also, it seems that sql fragments were not allowed by types (it's strange that no user have issues with this bug yet :thinking: but it's fixed now anyway)

I don't know if I should run the `build` script for transpilation; some commits where not transpiled, poluting the PR if I commit these changes :confused:

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -164,7 +164,7 @@ type Keys = string
 
 type SerializableObject<T, K extends readonly any[], TT> =
   number extends K['length'] ? {} :
-  (Record<Keys & (keyof T) & (K['length'] extends 0 ? string : K[number]), postgres.SerializableParameter<TT> | postgres.JSONValue> & Record<string, any>)
+  Partial<(Record<Keys & (keyof T) & (K['length'] extends 0 ? string : K[number]), postgres.ParameterOrJSON<TT> | undefined> & Record<string, any>)>
 
 type First<T, K extends readonly any[], TT> =
   // Tagged template string call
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -209,17 +209,17 @@ declare namespace postgres {
...
-    internal_position?: string;
-    internal_query?: string;
-    where?: string;
-    schema_name?: string;
-    table_name?: string;
-    column_name?: string;
-    data?: string;
-    type_name?: string;
-    constraint_name?: string;
+    detail?: string | undefined;
```
ignavan39, line 212: expression

```ts
field?: number
``` 
already implies that the type of field

```ts
number | undefined
```

types/index.d.ts
```diff
@@ -285,34 +285,34 @@ declare namespace postgres {
 
   interface Options<T extends PostgresTypeList> extends Partial<BaseOptions<T>> {
     /** @inheritdoc */
-    host?: string;
+    host?: string | undefined;
```
ignavan39, line 288: here is a similar question, why is this?

types/index.d.ts
```diff
@@ -437,21 +437,21 @@ declare namespace postgres {
 
   interface LargeObject {
     writable(options?: {
-      highWaterMark?: number,
-      start?: number
-    }): Promise<Writable>;
+      highWaterMark?: number | undefined,
+      start?: number | undefined
```
ignavan39, line 441: here is a similar question, why is this?

PR #459, opened 2022-08-09
Format code example for transform options

karlhorky:
Hi @porsager 👋

Quick PR to improve the formatting of the `transform` option code example


porsager:
Cool... don't know why it's wrapped in that anonymous function and process.exit thing - would you mind removing that now that you're at it ? :D

karlhorky:
Good point, done! 👍

PR #469, opened 2022-08-24
docs: add doc for json columns

pYassine:
Hi !
I recently encountered a problem updating a column in JSON format. Here is a small contribution to indicate it directly in the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,30 @@ const [{ a, b, c }] => await sql`
 `
 ```
 
+### Update jsonb columns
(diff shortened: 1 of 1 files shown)


jhonnyizidoro:
Thank you @pYassine! You already helped me even before the merge

porsager:
`sql.json()` is a remnant from v1, and not needed anymore, hence not in the docs.

There is an issue if your json value is not an object / array, but for that we have https://github.com/porsager/postgres/pull/392 in the works.

PR #480, opened 2022-09-15
Improve typing - Generic type should accept an object instead of an object[]

Newbie012:
Currently, in order to type a query, we need to pass it as generic as an array:

```ts
const q = await sql<{ text_col: string }[]>`select text_col from table_name`;
```

While it doesn't introduce any bugs, it's redundant to manually type it as an array (`type[]`) since it will always return an array.

I maintain a library ([SafeQL](https://github.com/ts-safeql/safeql)) that generates types based on queries (which uses this library, thanks to `.describe()`). While writing compatibility guides for each SQL library, I found out that this is the only library that requires adding `[]` at the end of the type.

Since it's a **breaking change**, I will fully understand if you would prefer not to merge it.

Although, there's a possibility to both support singles and arrays while maintaining backward compatibility, but it over complexes the type ([see demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEoHsDuyDeAoZQ5AbQXQBsBXAWxAC5kBnMKUAcwF0G4QBPAbjwBfPHhiUQCMMHQgmAR3IAeAkQAqyCAA9IIACaNkUCHD2zyvZAAp0AIwBWEKcgA+yCXogxQEPQEpiDldrO0dnNw8vH39kAF40LEC8AD4rSGoAB3I4SAY1CEzsyABlFnZGAEEoKDhePzzNHQh9Q1QTMxALKpreJQkAaxAsEGTkAH5kDTzA3FVCYzBKKDkcIWQ4Qx4BYVEyEGZkMABGOIVlHGRgPQYQGltoZCFkgANGCHInMEu9ZAAxVAA8gBZQ5wWwfAD6IDg1Agz0ERERyAA9MjkAA9MaiJFEPB7A5gABMp0YiiUFyuNzuDyEgRebw+ziuf0BILAYMh0NhzzmSNRGKxQA)):

```ts
<T extends readonly (object | undefined)[] | (object | undefined) = Row>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<unknown>)[]): PendingQuery<T extends ReadonlyArray<any> ? T : T[]>;
```

Note the new T:

```diff
- T extends readonly (object | undefined)[] = Row[]
+ T extends readonly (object | undefined)[] | (object | undefined) = Row[]
```

and the new return type:

```diff
- PendingQuery<T[]>;
+ PendingQuery<T extends ReadonlyArray<unknown> ? T : T[]>;
```

Results in:

image

On another note - There's a drift between the declaration types of nodejs and deno. I'm not sure if this is intentional.

### DIFF
--- a/deno/types/index.d.ts
+++ b/deno/types/index.d.ts
@@ -598,7 +598,7 @@ declare namespace postgres {
      * @param parameters Interpoled values of the template string
      * @returns A promise resolving to the result of your query
      */
-    <T extends readonly (object | undefined)[] = Row[]>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<any>)[]): PendingQuery<AsRowList<T>>;
+    <T extends (object | undefined) = Row>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<any>)[]): PendingQuery<AsRowList<T[]>>;
 
     CLOSE: {};
     END: this['CLOSE'];
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -596,7 +596,7 @@ declare namespace postgres {
      * @param parameters Interpoled values of the template string
      * @returns A promise resolving to the result of your query
      */
-    <T extends readonly (object | undefined)[] = Row[]>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<any>)[]): PendingQuery<T>;
+    <T extends (object | undefined) = Row>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<any>)[]): PendingQuery<T[]>;
 
     CLOSE: {};
     END: this['CLOSE'];


porsager:
I have a v4 coming up with other breaking changes, so if @Minigugus  thinks this is a good idea, i have no objections.

Minigugus:
@Newbie012 Initially, `sql<RowType>` was supported but was removed to only support `sql<RowType[]>` (see https://github.com/porsager/postgres/pull/84#issuecomment-647465191). The idea was to allows tuples and to kind of support multi-statement queries with `sql.unsafe`:

```ts
const [user, bug] = await sql<[User?]>`SELECT * FROM users WHERE email = ${email} LIMIT 1`;
// => error because `[User?]` only have 1 element but 2 requested

const [
  someTableRows,
  anotherTableRows
] = await sql<[SomeTableRow[], AnotherTableRow[]]>`
  SELECT * FROM someTable;
  SELECT * FROM anotherTable;
`;
```

(I haven't tested the *unsafe* example, @porsager is this still valid (multi-statements adds 1 level to the returned array)?)

Maybe it's time to reopen the debate, I guess it's up to the community now

Newbie012:
If I understand correctly:

1. `sql` supports multi stmts.
2. When there's a single stmt, the result should be `T[]`.
3. When there are two stmts, the result should be `[T[], R[]]`.
4. When there are three stmts, the results should by `[T[], R[], X[]], etc...

Then I assume [this](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEoHsDuyDeAoZQ5AbQXQBsBXAWxAC5kBnMKUAcwF0G4QBPAbjwBfPHjC8ADigAqlCeQgATAIJQocXgB5VUZBAAekEIsbJKIANYgsIYhwB8yALwEiOvYYjHTxUDGjIAGLAUMwANMgAdNF+AagQzByuREQA-CTBoWB2EdGRsvJKOhqa8cz2SSkpDO4GRia4yAogbGAAFgwADMhCyOl2yAzEOsSdHByCeDDmCGDA6CBMAI7kmsnI0h71PuZWNrnRu9aYtuPIAD5mlseLTmhYePYAFJDU8nCQDNIQb+QfEABlFjsRjFXgASi+Wy8DSONgG6QKChUahK0kcXwG+BSUAgYEoUEWOF6cFMPAEwlEZBAzGQYAAjM5lqscMhgIoGCAaAAjAJCewAA0YEAUszZiiCqAA8gBZOlwbkKAD6IDg1AgAsEVSIAHodcgAHqpUTawh4am03wmCKq9WMDhMxgrTTEVnszk8vkRVm2iAMZisFo9BxCkUQMXsyWy+WKiAqtUQfhMMNi31RuVgBXK30C9Z6w3GvDsxh4fNG0S+ktlwsWsB0gDMjudrvFHuovN0Qm9yF9-uBQa7jUQcwAbn7kNz0BQIDwhCHhaK65HAtKM1m476kwvw3W0yvo5nY-H1Unh8Axz2E+mY9mE7mUtW8EA) could work out?

image

porsager:
> (I haven't tested the unsafe example, @porsager is this still valid (multi-statements adds 1 level to the returned array)?)

unsafe queries without any arguments are implicitly sent as simple queries allowing for multiple statements. This is also possible with the next release using tagged template literals by calling `simple` like this: ``` sql`select 1;select 2`.simple() ``` . I am considering changing unsafe in v4 to require using `simple()`.

porsager:
@karlhorky could you maybe help figure out if this is something that should be merged?

Newbie012:
While I still think my suggested approach makes a bit more sense, I would advise not going for it since the added value here is not worth the breaking change.

Feel free to close this PR 🙂

karlhorky:
I think the main value is this, from the original PR description:

> While writing compatibility guides for each SQL library, I found out that this is the only library that requires adding `[]` at the end of the type.

If consistency with other libraries in the ecosystem is important for Postgres.js, then maybe the approach of accepting an object instead of an array or tuple would be better as the only API for generic parameters (as a breaking change for v4).

One other downside not discussed explicitly above is the other benefits that can come from using tuples as the generic parameters data type, as @yckao and @Minigugus [mentioned in various #84 comments](https://github.com/porsager/postgres/pull/84#issuecomment-647046073). In our team we switched away from using tuples because [SafeQL does not support the tuple notation](https://github.com/ts-safeql/safeql/issues/58) and [TypeScript can be made to check for indexed access with `noUncheckedIndexedAccess`](https://github.com/ts-safeql/safeql/issues/58#issuecomment-1258752196). So potentially tuples are less useful because of this.

So I would suggest one of two paths:

1. If consistency with other library generic parameters is important to Postgres.js, then switch to object-only (remove array and tuple support) in v4
2. If it's not very important, or not important now, then close the PR

porsager:
Thanks to both of you for the follow up and explanations. 


How about the 3rd option? 

> Although, there's a possibility to both support singles and arrays while maintaining backward compatibility, but it over complexes the type ([see demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEoHsDuyDeAoZQ5AbQXQBsBXAWxAC5kBnMKUAcwF0G4QBPAbjwBfPHhiUQCMMHQgmAR3IAeAkQAqyCAA9IIACaNkUCHD2zyvZAAp0AIwBWEKcgA+yCXogxQEPQEpiDldrO0dnNw8vH39kAF40LEC8AD4rSGoAB3I4SAY1CEzsyABlFnZGAEEoKDhePzzNHQh9Q1QTMxALKpreJQkAaxAsEGTkAH5kDTzA3FVCYzBKKDkcIWQ4Qx4BYVEyEGZkMABGOIVlHGRgPQYQGltoZCFkgANGCHInMEu9ZAAxVAA8gBZQ5wWwfAD6IDg1Agz0ERERyAA9MjkAA9MaiJFEPB7A5gABMp0YiiUFyuNzuDyEgRebw+ziuf0BILAYMh0NhzzmSNRGKxQA)):

```
<T extends readonly (object | undefined)[] | (object | undefined) = Row>(template: TemplateStringsArray, ...parameters: readonly (SerializableParameter<TTypes[keyof TTypes]> | PendingQuery<unknown>)[]): PendingQuery<T extends ReadonlyArray<any> ? T : T[]>;
```

What are the downsides of this?

porsager:
About typescript decisions in this project, i rely on @karlhorky and @Minigugus to guide it, if it doesn't affect the js API 😉

karlhorky:
If I'm understanding correctly, the third option above is to support both arrays and objects?

I guess I would suggest avoiding this, seems like a bigger downside (more complexity in implementing and maintaining it) without too much upside (allowing object generic arguments seems mainly good for compatibility, and I guess also supporting arrays would be for backwards compatibility). The combined benefits of library compatibility and backwards compatibility don't seem to outweigh the downside of the increased complexity - for me at least.

So I would probably still stay with [option 1 or 2](https://github.com/porsager/postgres/pull/480#issuecomment-1627645870) - could go either way on which one, I think they have close to equal upsides and downsides.

Maybe @Minigugus has a different opinion.

porsager:
Ok :) Thanks a lot... Let's close this then, and perhaps revisit later if @Minigugus thinks it's worth considering. (haven't heard from him in a while).

PR #492, opened 2022-10-03
Docs for multiple updates in one query

PierBover:
Like we chatted on Gitter, I've added this bit of information.

I tried to maintain the style of the code and SQL consistent with the other examples. Hope the location is ok, I didn't know where to put it.


porsager:
Very nice @PierBover ! Thank you !

PR #494, opened 2022-10-04
add compatible case for default transforms

the-red-herring:
Explain that snake_case is the only case for column names that is compatible with the default transform functions.


porsager:
Nice to have clarified! Thank you 👏

PR #497, opened 2022-10-06
Fix connection uri encoding (Issue #399)

### LINKED ISSUE #399: postgres(url) wrongly encodes password field when password contains special characters
PROBLEM

Calling postgres(url) wrongly encodes the password field when password contain special characters. However calling postgres(obj) with the same credentials works as expected.

CONTEXT

I was getting an auth error when trying to connect to a database using a connection string that had a special character `^` in the password field. I had checked the credentials were correct by using psql. So credentials are right but can't connect, being a first-time user of the library, I thought was probably doing something stupid. I then tried using a connection object with the same credentials and things just worked as expected.

WAT?

I then tried passing the same connection string but overriding the password field as suggested in the docs, and that worked as well. Looking at the difference of the resulting object from calling `postgres()` I noticed that the password has some extra funny characters in it when I used the connection string. Here is some code to illustrates the difference:

```js
const postgres = require("postgres");
const url = "postgres://user:pass^word@localhost";
const a = postgres(url).options.pass;
const b = postgres({ user: "user", pass: "pass^word" }).options.pass;
console.log(a);
console.log(b);
console.log(a == b);
```

Output:

```
pass%5Eword
pass^word
false
```

WAT?

Why are they different? Could it be that library encodes the password when we pass an url but it doesn't we use an object?

It seems so:

https://github.com/porsager/postgres/blob/50e6baf197b09a4b29461c0ca18b323bd601925a/cjs/src/index.js#L483

Now, I'm not 100% sure if this is a bug or not, but this behaviour was definitely unexpected to me. I skimmed the postgres documentation and they say that connection strings should be url encoded, however because the library is parsing the url and using an object before it sends the credentials anyway, maybe the library should decode the user and password before attempting to connect?

POTENTIAL FIX:

Use `decodeURIComponent` to decode the user and pass (at least) after parsing the url:

```js
// postgres/cjs/src/index.js
function parseUrl(url) {
  if (typeof url !== 'string')
    return { url: { searchParams: new Map() } }

  let host = url
  host = host.slice(host.indexOf('://') + 3).split(/[?/]/)[0]
  host = decodeURIComponent(host.slice(host.indexOf('@') + 1))

  const urlObj = new URL(url.replace(host, host.split(',')[0]));

  return {
    url: {
      username: decodeURIComponent(urlObj.username),
      password: decodeURIComponent(urlObj.password),
      host: urlObj.host,
      hostname: urlObj.hostname,
      port: urlObj.port,
      pathname: urlObj.pathname,
      searchParams: urlObj.searchParams,
    },
    multihost: host.indexOf(',') > -1 && host
  }
}
```

Thoughts?

I hope it's just me still doing something stupid :)

vnava-nsw:
Use decodeURIComponent to decode the user and pass from connection uri.
So that connections to pg work when using uri strings with special characters in user or pass.

Fixes #399

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -486,8 +486,18 @@ function parseUrl(url) {
   host = host.slice(host.indexOf('://') + 3).split(/[?/]/)[0]
   host = decodeURIComponent(host.slice(host.indexOf('@') + 1))
 
+  const urlObj = new URL(url.replace(host, host.split(',')[0]))
+
   return {
-    url: new URL(url.replace(host, host.split(',')[0])),
+    url: {
+      username: decodeURIComponent(urlObj.username),
+      password: decodeURIComponent(urlObj.password),
+      host: urlObj.host,
+      hostname: urlObj.hostname,
+      port: urlObj.port,
+      pathname: urlObj.pathname,
+      searchParams: urlObj.searchParams
+    },
     multihost: host.indexOf(',') > -1 && host
   }
 }
--- a/tests/index.js
+++ b/tests/index.js
@@ -351,6 +351,11 @@ t('Connect using uri', async() =>
   })]
 )
 
+t('Options from uri with special characters in user and pass', async() => {
+  const opt = postgres({ user: 'öla', pass: 'pass^word' }).options
+  return [[opt.user, opt.pass].toString(), 'öla,pass^word']
+})
+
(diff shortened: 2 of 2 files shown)


porsager:
Very nice @vnava-nsw , and sorry for the late reply. looks good!

PR #504, opened 2022-10-13
Doc improvements

ChristophP:
- add a quick overview over all the ways to do interpolation and when to use them 
grafik

- remove the documented arguments from the `.describe()` method as it doesn't seem to accept any https://github.com/porsager/postgres/blob/master/src/query.js#L114

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,16 +339,16 @@ sql`
 select "id" from "users"
 ```
 
-### Quick Primer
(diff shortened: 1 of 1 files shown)


porsager:
Awesome @ChristophP ! And good catch on the describe doc error. Looks like copy pasta from the cursor section :P

PR #505, opened 2022-10-14
Allow for incomplete custom types

acarstoiu:
Specifically, I'd like to be able to get away with a type definition consisting in just the `to` property, like this:
`
    int4: {
      to: 23
    }
`
That's because I merely want to use a name for a PostgreSQL type OID and the usual conversions are already defined in this file. As a fallback, the default serialization in _src/connection.js_ at line _912_ is just fine.


porsager:
Looks good - I can see the purpose of that!

PR #508, opened 2022-10-18
try to connect to multiple hosts + some debug output

furiozo-ga:
Hello, I really liked the features of your postgres.js module and decided I am going to use it for its support for multiple hosts. I intend to use it for a high availability setup.
But unfortunately I immediately encountered a problem : if the first host is down, the driver doesn't try to connect to the other hosts, just gives up.
I really do not understand how the module as a whole works, but with my limited understanding I made some modifications that allow it to connect if there is at least one working host.
This is probably not enough for it to be fully functional and you will have to rework it but that's as far as my skill goes.
I hope this functionality gets in !

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -103,6 +103,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     , nonce = null
     , query = null
     , final = null
+    , firstFailedHostIndex = null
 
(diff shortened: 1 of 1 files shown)


porsager:
Please reopen an issue if you still have issues with multiple hosts and the latest version

PR #510, opened 2022-10-20
Fix null in nested json array transform error

Eprince-hub:
Transforming a json with `null` value throws the below error
```
 Uncaught TypeError: Cannot convert undefined or null to object
```
For example, the object below throws an error when it hits the  `Object.entries(x)` part of the transform function
```js
obj = {
id: 1,
name: 'Victor,
age: null
}
```
Screenshot 2022-10-19 at 10 14 15

I think checking for `x` before passing it to the `Object.entries(x)` would prevent this error. So we only pass `x` to be transformed when we are sure that it is `object`,  `array`,  not `null` and `json`. 
The diff below would be the suggested fix
```diff
- column.type === 114 || column.type === 3802
+ (x && typeof x in allowList) && (column.type === 114 || column.type === 3802)
```

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -39,6 +39,10 @@ export const types = {
   }
 }
 
+const allowList = {
+  object: typeof Object
+}
+
 class NotTagged { then() { notTagged() } catch() { notTagged() } finally() { notTagged() }}
 
 export class Identifier extends NotTagged {
@@ -329,7 +333,7 @@ export const fromKebab = x => x.replace(/-/g, '_')
 
 function createJsonTransform(fn) {
   return function jsonTransform(x, column) {
-    return column.type === 114 || column.type === 3802
+    return (x && typeof x in allowList) && (column.type === 114 || column.type === 3802)
       ? Array.isArray(x)
         ? x.map(x => jsonTransform(x, column))
         : Object.entries(x).reduce((acc, [k, v]) => Object.assign(acc, { [fn(k)]: v }), {})
--- a/src/types.js
+++ b/src/types.js
@@ -39,10 +39,6 @@ export const types = {
   }
 }
 
-const allowList = {
-  object: typeof Object
-}
-
 class NotTagged { then() { notTagged() } catch() { notTagged() } finally() { notTagged() }}
 
 export class Identifier extends NotTagged {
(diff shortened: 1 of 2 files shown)


src/types.js
```diff
@@ -329,7 +333,7 @@ export const fromKebab = x => x.replace(/-/g, '_')
 
 function createJsonTransform(fn) {
   return function jsonTransform(x, column) {
-    return column.type === 114 || column.type === 3802
+    return (x && typeof x in allowList) && (column.type === 114 || column.type === 3802)
```
karlhorky, line 336: Instead of this extra `allowList` variable, you can probably just do the check inline.

Also, you may want to instead try checking against:

- Is `x` an array
- Is `x` an object
- Is `x` not `null`

It's possible that you don't need 3 different conditions for this though...

Eprince-hub: Commit
https://github.com/porsager/postgres/pull/510/commits/b0b2e0cca3f0ab0becc42d0cb689fdd1fa0e5f6e

karlhorky: seems better to me 👍 

```js
return typeof x === 'object' && x !== null
```

@porsager condition good for you too?

tests/index.js
```diff
@@ -611,6 +611,14 @@ t('Transform nested json in arrays', async() => {
   return ['aBcD', (await sql`select '[{"a_b":1},{"c_d":2}]'::jsonb as x`)[0].x.map(Object.keys).join('')]
 })
 
+t('Transform null json in arrays', async() => {
+  const sql = postgres({
+    ...options,
+    transform: postgres.camel
+  })
+  return [null, (await sql`select '[{"a_b":null},{"c_d":null}]'::jsonb as x`)[0].x.map(Object.keys).join('')]
```
karlhorky, line 619: I'm not sure this condition does what you think it does (I imagine this test will fail)

this `.map(Object.keys)` and `.join()` don't seem like the right thing to do here in this test - you don't care about the object keys, you care about the values

karlhorky, line 619: I'm not sure this condition does what you think it does (I imagine this test will fail)

setting a jsonb value to an array has already been done in the previous test - your new test should make sure that json and jsonb values which are both not arrays and also not objects do not get transformed (and do not cause this crash that you experienced)

Eprince-hub: The commit
https://github.com/porsager/postgres/pull/510/commits/b8242f357053fa006bc2af394cb55db2bcc7fc05

Eprince-hub: @porsager, What do you think about the PR? 
I think it is ready for you

karlhorky:
@porsager any thoughts on the approach in this PR here? Would be happy to help to get this merged, in case there's anything left to implement.

porsager:
Hey.. thanks for the ping.. I've simply been too busy lately, so haven't checked it out, but looking now I think that's perfectly fine - it's how I would have done it ;)

Thanks!

src/types.js
porsager: Yup - perfect!

Eprince-hub:
Thanks a lot for the merge 👍
Could we be expecting this to be released as `postgres@3.3.2`?

PR #531, opened 2022-12-10
Fix writing host and port on connection timeouts

dirkdev98:
Received a `write CONNECT_TIMEOUT undefined:undefined` when the configuration was correct, but something in the network was incorrectly configured.

The Deno polyfill already sets these options, but the Node.js socket doesn't expose them. Defaulting to the options object would also be an option here, but then we need to know which `options.host[idx]` is currently used.


porsager:
Ah, great - looks good - thank you!

Issue #471, opened 2022-08-25
Boolean array type inference error


uasan:
Boolean array, always passed as a boolean value, here is an example:
```JS
await sql`SELECT ${[false, true]} AS bool_array`
// Result(1) [ { bool_array: false } ]
```
this query is throwing an error
```JS
await sql`SELECT ${[true, false]}::bool[]`;
// Error cannot cast type boolean to boolean[]
```

binikingi:
For now what im doing is this:
```
await sql`SELECT ${[false, false].map(x => x ? 1 : 0)}::INT[]::BOOLEAN[]`
```

Ciantic:
> PostgresError: column "values" is of type boolean[] but expression is of type boolean

This happens with unsafe too:

```typescript
    await sql.unsafe(`CREATE TABLE test_bool_arr (values boolean[]);`).execute();

     // PostgresError: column "values" is of type boolean[] but expression is of type boolean
     await sql`INSERT INTO test_bool_arr (values) VALUES (${[true, false, true]})`.execute();

    // PostgresError: column "values" is of type boolean[] but expression is of type boolean
    await sql.unsafe(`INSERT INTO test_bool_arr (values) VALUES ($1)`, [[true, false, true],]).execute();
```

PR #533, opened 2022-12-14
fix array type inference

Newbie012:
fixes https://github.com/porsager/postgres/issues/471

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -213,6 +213,17 @@ export const escapeIdentifier = function escape(str) {
   return '"' + str.replace(/"/g, '""').replace(/\./g, '"."') + '"'
 }
 
+const defaultArrayTypes = {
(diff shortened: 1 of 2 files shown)


porsager:
Doesn't https://github.com/porsager/postgres/pull/392 fix this?

Newbie012:
Seems like it. I ran my tests against your branch and they passed. Although, this PR seems like a "less" breaking change rather than #392. If this PR would be merged, then it should probably be reverted when #392 is out.
Feel free to close this PR if it's not relevant.

porsager:
Great :)

Indeed it's less breaking, but it's still breaking enough, like my PR, so that I feel it would require a major bump. I'm unfortunately too wound up to move on anything as it is, but hopefully I'll find time during Christmas to catch up on it all.

PR #543, opened 2023-01-16
pass servername to tls.connect

antonyc:
This addresses the inability to connect using SNI to a TCP proxy hosting multiple Postgres instances behind itself. This is, for example, the case of https://neon.tech. With this patch, it is possible to establish a connection to the Neon compute-node by specifying `postgres('database_url', { ssl: true })` while connecting from Javascript. Without this patch, the `servername` is not passed.

See [nodejs docs](https://nodejs.org/docs/latest/api/tls.html#tls_tls_connect_options_callback) for details on `servername` option and SNI extension to TLS.

### DIFF
--- a/cjs/src/connection.js
+++ b/cjs/src/connection.js
@@ -263,18 +263,22 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     if (!canSSL && ssl === 'prefer')
       return connected()
 
-    socket.removeAllListeners()
-    socket = tls.connect({
+    const sslOptions = {
       socket,
-      ...(ssl === 'require' || ssl === 'allow' || ssl === 'prefer'
-        ? { rejectUnauthorized: false }
-        : ssl === 'verify-full'
-          ? {}
-          : typeof ssl === 'object'
-            ? ssl
-            : {}
-      )
-    })
+    }
+
+    if (typeof ssl === 'object') {
+      Object.assign(sslOptions, ssl)
+    } else if (['require', 'allow', 'prefer'].includes(ssl)) {
+      sslOptions.rejectUnauthorized = false
+    }
+
+    if (!sslOptions.servername && net.isIP(host) === 0) {
+      sslOptions.servername = host[0]
+    }
+
+    socket.removeAllListeners()
+    socket = tls.connect(sslOptions)
     socket.on('secureConnect', connected)
     socket.on('error', error)
     socket.on('close', closed)
--- a/cjs/src/connection.js
+++ b/cjs/src/connection.js
@@ -256,6 +256,18 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     socket.destroy()
   }
 
+  function isIP(input) {
+    if (net.isIP) {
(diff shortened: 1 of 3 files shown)


antonyc:
@porsager , please have a look.

antonyc:
@porsager I saw that tests exist in this repo, but they are run manually. And saw test run was unsuccessful. I tried to run them locally with act from nektos, but failed. So I think I fixed the test but waiting for you to run them. Sorry for pinging/

porsager:
There u go 😉 I don't think I can change that I need to manually allow the very first test run for a user, but now they should at least run for you 😉

bonesoul:
also looking for this to get working with neon.tech

antonyc:
@porsager still cannot run the workflow, but updated the tests. Please have a look.

antonyc:
Oh, the tests passed somehow. I did not trigger the workflow.
Can we squash & merge it?

porsager:
Great, perhaps we can have @bonesoul give your branch a shot to see if it works before I look closer + merge?

antonyc:
@bonesoul could you give it a try?

bonesoul:
well I dropped neon.tech support on my project cause of the issue.

porsager:
Hi @antonyc .. Finally had some time to check this out and give neon.tech a try.. I'm curious I think simply doing the following solves the issue with minimal changes:

```diff
     socket = tls.connect({
       socket,
+      servername: socket.host,
       ...(ssl === 'require' || ssl === 'allow' || ssl === 'prefer'
         ? { rejectUnauthorized: false }
```

porsager:
Oh right - it also seems to generate a pretty error in node complaining if it's an IP. 

Ok, so since `socket.host` contains the current host being connected to, this should be enough right?

```diff
     socket = tls.connect({
       socket,
+      servername: net.isIp(socket.host) ? undefind : socket.host,
       ...(ssl === 'require' || ssl === 'allow' || ssl === 'prefer'
         ? { rejectUnauthorized: false }
```

To include deno support, the following could be added to polyfill.js instead of mixing platform specifics into the main src
```diff
  import { Buffer } from 'https://deno.land/std@0.132.0/node/buffer.ts'
+ import { isIP } from 'https://deno.land/std@0.132.0/node/net.ts'

  const events = () => ({ data: [], error: [], drain: [], connect: [], secureConnect: [], close: [] })

  export const net = {
+   isIP,
    createServer() {
      const server =  {
```

Would also be nice if you could remove the artifacts from cjs and deno builds in the PR.

boromisp:
FYI there are workarounds for neon.tech: https://neon.tech/docs/connect/connectivity-issues, some of them would probably work with postgres.js as-is. Nonetheless, having SNI work will be nicer.

porsager:
I'm gonna add my suggestion above unless @antonyc has any objections or wants to edit his PR to match it.

antonyc:
Looks good to me, no expert on deno

porsager:
Cool.. want me to push it or do you wanna edit this pr?

antonyc:
Thanks!

Issue #576, opened 2023-04-05
_box (box[]) isn't formatted/parsed as an array correctly


Bas950:
https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT
![image](https://user-images.githubusercontent.com/34792838/186644918-e4780f25-3d34-4454-a7cf-2f879461ab8f.png)
I am guessing since box has a different delimiter character the formatter and parser are messing up.

PR #578, opened 2023-04-05
feat: don't override array type if already exists

### LINKED ISSUE #577: When a custom array parser is specified it still uses it's own
Specifying a custom array parser in options.types is just completely ignored.

For example using:
```ts
{
  to: 719, //719 = _circle (circle[])
  from: [719],
  serialize: () => {
    //...
   },
  parse: () => { 
    //... 
  },
}
```

Bas950:
Resolves #577

### DIFF
--- a/cjs/src/connection.js
+++ b/cjs/src/connection.js
@@ -735,6 +735,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function addArrayType(oid, typarray) {
+    if (!!options.parsers[typarray] && !!options.serializers[typarray]) return;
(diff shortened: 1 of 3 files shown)


Bas950:
> Looks great - Thanks a lot!

Yea no problem, found a few more errors, looking into those at the moment.
* #576 
* If you pass a custom OID parser it also gets used for internal functions which breaks a few things...

Bas950:
@porsager Any release planned by any chance?

porsager:
nope, but I'll make one today 👍

Bas950:
Thank you!

PR #585, opened 2023-04-12
readme: add correction in the "Multiple updates in one query" section

paulovieira:
In the "Multiple updates in one query" it seems it is necessary to do explicit type casting if the data type is not text (in both the SET and WHERE clauses). 

If not, I see one of these errors:

- `column "sort_order" is of type integer but expression is of type text`
- `operator does not exist: integer = text`

I also added the "returning" at the end to confirm that the data was updated.


porsager:
Oh cool. Thanks a lot :)

PR #593, opened 2023-05-08
Feat: add role options to transactions

guilherme-miyake:
I really wanted to be able to set the transaction role before running the queries proper, for instance in my use case, I intend to only expose an transaction with the role already set, so even if someone uses admin credentials to start the application, it will still run all commands using the proper permissions.

It maintains compatibility with `sql.begin` and prepares for further extension.
P.s.: Some changes on language dirs are from previous commits


porsager:
I don't think that belongs in this libary.

just do 
```js
await sql.begin(sql => [
  sql`set role ${ sql('postgres_js_test_set_role') }`,
  sql`create table test (x int)`
])
```

PR #599, opened 2023-05-17
Support for Cloudflare Workers

porsager:
Initial support for using Postgres.js in Cloudflare Workers (tcp support was just released today).


superboss224:
Hello !
Is the update with cloudflare workers already done ?

Thank you for your fast update !

daniel-xyz:
That's awesome. Hopefully that means it'll soon work on Vercel Edge Functions too, since they're built on top of Cloudflare Workers as far as I know. 👀

porsager:
Ok, I think we're about there with the great work from @wackfx !

It would be nice to have a small section in the README.md about it as well, if anyone is up for that? Perhaps a separate PR, so we can get this merged.

Issue #217, opened 2021-08-24
feature request: add on conflict set to dynamic insert query helper


alextes:
I have a query like so:
```ts
const insertAnalysisT = () => sql`
	  INSERT INTO base_fees_per_block (
		    hash,
		    number
	  )
	  VALUES (
		    ${block.hash},
		    ${block.number}
	  )
	  ON CONFLICT (number) DO UPDATE
	  SET
		    hash = ${block.hash},
		    number = ${block.number}
`;
```

Because of the conflict but I can't use the dynamic query helper. A helper that simply sets all columns that are inserted on a conflict with a specifiable list of columns (in this case just `number` but could be multiple), would be quite flexible and for me helpful. This query is a lot longer in actuality.

If there are others (👍 ) it may be interesting to implement!

antonbrams:
unfortunatelly it doesn't work for me

```
const key = 'id'
const table = 'orders_status'
const rows = [
	{id: 2, title: 'test1'},
	{id: 3, title: 'test2'},
]
sql`
	INSERT INTO ${sql(table)} ${sql(rows)}
	ON CONFLICT (${key}) DO UPDATE
	SET ${sql(rows)}
`
```

even if i hard code the conflict id, it throws an error
"PostgresError: column "0" of relation "orders_status" does not exist"

what am i doing wrong?

alextes:
I'm not sure but I think the issue is with the conflict clause:

```
ON CONFLICT (${key}) DO UPDATE
```

you're supposed to provide the name of the column, I believe for column names the syntax would be 
```js
`${sql(key)}`
```

Maybe. You can try `ON CONFLICT (id) DO UPDATE` to confirm that's what is going wrong. You can also use the `debug` option to see what SQL is actually being sent which might make the problem obvious. You can see a short description of it here: https://github.com/porsager/postgres#all-postgres-options

porsager:
You're missing sql() around key 😊

antonbrams:
> You're missing sql() around key 😊

i've tested it too... but please read carefully, what i've wrote - even this configuration

```
sql`
	INSERT INTO ${sql(table)} ${sql(rows)}
	ON CONFLICT (id) DO UPDATE
	SET ${sql(rows)}
`
```
throws an error `"PostgresError: column "0" of relation "orders_status" does not exist"`

porsager:
Sorry. Replied on mobile, so missed the subsequent replies before writing. 

What version are you on?

antonbrams:
[...]

PR #613, opened 2023-06-11
Allow array of Fragments in ParameterOrFragment

karlhorky:
Ref: https://github.com/porsager/postgres/issues/217#issuecomment-1586176144

cc @Minigugus does this look right? It should fix the error that I showed in the comment above


karlhorky:
Now released in [`postgres@3.4.0`](https://github.com/porsager/postgres/releases/tag/v3.4.0), confirmed as working, thanks @porsager ! 🙌

PR #621, opened 2023-06-26
README.md - improve the "Multiple statements in one query" section

paulovieira:
- add links for the official documentation
- escape the backtick character
- change the subtitle to "await sql``.simple()" instead of "await sql`select 1; select 2;`.simple()" (to be coherent with the other subtitles)
- add a small example below

before

Screenshot_2023-06-26_11-16-08-before

after

Screenshot_2023-06-26_11-15-58-after


porsager:
Thanks a lot - much better

PR #628, opened 2023-07-02
create beginPrepared function

shayan-shojaei:
Added a new function called `beginPrepared` to support prepared transactions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -637,6 +637,26 @@ sql.begin('read write', async sql => {
 })
 ```
 
+
+#### PREPARE `await sql.prepare([name]) -> fn()`
+
+Indicates that the transactions should be prepared using the `PREPARED TRANASCTION [NAME]` statement
+instead of being committed.
+
+```js
+sql.begin('read write', async sql => {
+  const [user] = await sql`
+    insert into users (
+      name
+    ) values (
+      'Murray'
+    )
+  `
+    
+  await sql.prepare('tx1')
+})
+```
+
 Do note that you can often achieve the same result using [`WITH` queries (Common Table Expressions)](https://www.postgresql.org/docs/current/queries-with.html) instead of using transactions.
 
 ## Data Transformation
--- a/src/index.js
+++ b/src/index.js
@@ -77,7 +77,6 @@ function Postgres(a, b) {
     reserve,
     listen,
     begin,
-    beginPrepared,
     close,
     end
   })
@@ -231,11 +230,12 @@ function Postgres(a, b) {
     }
   }
 
-  async function begin(options, fn, transactionId) {
+  async function begin(options, fn) {
     !fn && (fn = options, options = '')
     const queries = Queue()
     let savepoints = 0
       , connection
+    let transactionId = null
 
     try {
       await sql.unsafe('begin ' + options.replace(/[^a-z ]/ig, ''), [], { onexecute }).execute()
@@ -247,6 +247,7 @@ function Postgres(a, b) {
     async function scope(c, fn, name) {
       const sql = Sql(handler)
       sql.savepoint = savepoint
+      sql.prepare = prepare
       let uncaughtError
         , result
 
@@ -290,6 +291,9 @@ function Postgres(a, b) {
       }
     }
 
+    async function prepare(name) {
+      transactionId = name
+    }
     function onexecute(c) {
       connection = c
       move(c, reserved)
@@ -299,9 +303,6 @@ function Postgres(a, b) {
     }
   }
 
-  async function beginPrepared(transactionId, options, fn) {
-    return begin(options, fn, transactionId)
-  }
 
   function move(c, queue) {
     c.queue.remove(c)
--- a/src/index.js
+++ b/src/index.js
@@ -77,6 +77,7 @@ function Postgres(a, b) {
     reserve,
     listen,
     begin,
+    beginPrepared,
     close,
     end
   })
@@ -230,7 +231,7 @@ function Postgres(a, b) {
     }
   }
 
-  async function begin(options, fn) {
+  async function begin(options, fn, transactionId) {
     !fn && (fn = options, options = '')
     const queries = Queue()
     let savepoints = 0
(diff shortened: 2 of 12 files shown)


porsager:
My idea was that .prepare() would be available on the sql transaction instance, so you would still use .begin()

Like this:

```js
await sql.begin(async sql => {
  ...queries
  await sql.prepare(name)
})
```

shayan-shojaei:
This would have been a great option if running other queries after the `prepare` function was an option, but considering that it's not, I would imagine this would be a more elegant option.

porsager:
There's nothing preventing us from implementing it such that you can? sql.prepare(name) can simply be a function that caches the name and signals that instead of commit we should do prepare transaction name

shayan-shojaei:
So your idea is that `sql.prepare(name)` does not instantly prepare the transaction? Instead it signals that the transaction should end with a prepared transaction?

porsager:
The cool thing is that you'll also be able to conditionally decide if you want to prepare or commit the transaction depending on the result of queries in the transaction.

porsager:
> So your idea is that `sql.prepare(name)` does not instantly prepare the transaction? Instead it signals that the transaction should end with a prepared transaction?

Exactly.

shayan-shojaei:
Yes that would be nice, I'll implement it like so.

porsager:
Cool.. Mind adding tests and documentation for this new method?

shayan-shojaei:
> Cool.. Mind adding tests and documentation for this new method?

done

porsager:
Looks good so far. Perhaps the test should open another connection and try to commit the prepared transaction?

shayan-shojaei:
> Looks good so far. Perhaps the test should open another connection and try to commit the prepared transaction?

Yeah I just thought of that. See if the current test would suffice.

porsager:
🎉 perfect.. Seems we need to enable prepared transactions in the bootstrap script for tests to work. 

Also - very curious what scenario you're in needing this feature :) sounds exotic ;)

porsager:
Add this line to bootstrap.js
`exec('psql', ['-c', 'alter system set max_prepared_transactions to 100'])`

shayan-shojaei:
> 🎉 perfect.. Seems we need to enable prepared transactions in the bootstrap script for tests to work.
> 
> Also - very curious what scenario you're in - needing this feature yum sounds exotic desert_island

We're trying to implement distributed transactions using the 2PC method. =)

shayan-shojaei:
> Add this line to bootstrap.js `exec('psql', ['-c', 'alter system set max_prepared_transactions to 10'])`

I don't think this has enabled them.

porsager:
No, seems it can only be set at start, so we need to edit postgresql.conf and restart 😐

shayan-shojaei:
> No, seems it can only be set at start, so we need to edit postgresql.conf and restart neutral_face

I think that could be handled if a line is added to the `test.yml` file to add that line at start.

porsager:
I'd like an idempotent solution in the bootstrap.js script since I run tests most locally (and that ought to fix github as well). I'm looking into a solution ;)

porsager:
I remember why i hated this :P Doing a cross platform way of restarting the server is quite annoying 😅 I think for now saying you have to do this locally yourself is ok, and the following line to replace your's in the .yml ought to do the trick:

`sudo sed -i 's/.*max_prepared_transactions.*/max_prepared_transactions = 100/' /etc/postgresql/${{ matrix.postgres }}/main/postgresql.conf`

shayan-shojaei:
> I remember why i hated this now 😮‍💨 Doing a cross platform way of restarting the server is quite annoying 😅 I think for now saying you have to do this locally yourself is ok, and the following line to replace your's in the .yml ought to do the trick for github actions.:
> 
> `sudo sed -i 's/.*max_prepared_transactions.*/max_prepared_transactions = 100/' /etc/postgresql/${{ matrix.postgres }}/main/postgresql.conf`

Yeah I can't think of a pretty way of doing that either.

porsager:
Everything looks good now 👏

types/index.d.ts
```diff
@@ -696,6 +696,8 @@ declare namespace postgres {
   interface TransactionSql<TTypes extends Record<string, unknown> = {}> extends Sql<TTypes> {
     savepoint<T>(cb: (sql: TransactionSql<TTypes>) => T | Promise<T>): Promise<UnwrapPromiseArray<T>>;
     savepoint<T>(name: string, cb: (sql: TransactionSql<TTypes>) => T | Promise<T>): Promise<UnwrapPromiseArray<T>>;
+
+    prepare<T>(name: string): Promise<UnwrapPromiseArray<T>>;
```
Minigugus, line 700: The `T` parameter is completely useless here (not used in parameters or in return type) so it should be removed to avoid confusing users

```suggestion
    prepare(name: string): Promise<void>;
```

src/index.js
```diff
@@ -266,7 +268,11 @@ function Postgres(a, b) {
         throw e instanceof PostgresError && e.code === '25P02' && uncaughtError || e
       }
 
-      !name && await sql`commit`
+      if (transactionId) {
+        !name && await sql.unsafe(`prepare transaction '${transactionId}'`)
```
Minigugus, line 272: `transactionId` isn't escaped so users code could easily break with a `transactionId` that contains a `'`. Also, if for some reason a user pass a unsanitized value as input, there's a risk of SQL injection :confused:

README.md
```diff
@@ -637,6 +637,26 @@ sql.begin('read write', async sql => {
 })
 ```
 
+
+#### PREPARE `await sql.prepare([name]) -> fn()`
+
+Indicates that the transactions should be prepared using the `PREPARED TRANASCTION [NAME]` statement
```
Minigugus, line 643: Typo

```suggestion
Indicates that the transactions should be prepared using the `PREPARED TRANSACTION [NAME]` statement
```

Minigugus (commented):
Sorry I'm a month late... :sweat: Just to report some issues that might be worth to fix (it's still on master as of now)

PR #637, opened 2023-07-14
Update README.md (prepared transactions)

paulovieira:
- correct typo
- add link to the official docs
- change the subsection name to "PREPARE TRANSACTION" instead of "PREPARE" (because "PREPARE" is more associated with "prepared statements")

One thing that is still a bit confusing in this section is the final sentence "Do note that you can often achieve...". It seems like it is referring to the "PREPARE" subsection, but in reality it is referring to the initial "BEGIN / COMMIT" subsection, no?

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ const user = {
   age: 68
 }
 
-sql`
+await sql`
   insert into users ${
     sql(user, 'name', 'age')
   }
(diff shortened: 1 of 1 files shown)


porsager:
Thanks ;)

You're right about the final sentence, that wasn't down there originally, so might have been pushed down by a PR :P You're welcome to move it to it's proper place in this PR too.

porsager:
Awesome, but I'd rather not include the stackoverflow example.

paulovieira:
- Removed link to stackoverflow

porsager:
Thanks a lot

PR #651, opened 2023-07-27
Fixed typo in README.md

miguelvictor:
Fixed typos and improved word ordering for better coherence.


thomas992:
`It's possible to create multiple updates in a single query. It's necessary to use arrays instead of objects to ensure that the order of the items correspond with the column names.`

I would go even farther and say

`To create multiple updates in a single query it is necessary to use arrays instead of objects to ensure that the order of the items correspond with the column names.`

porsager:
lgtm - thanks

Issue #649, opened 2023-07-21
Transactions example in the README.md doesn't work as expected


MarisaCodes:
[Transactions section in README](https://github.com/porsager/postgres#transactions)

```js
const [user, account] = await sql.begin(async sql => {
  const [user] = await sql`
    insert into users (
      name
    ) values (
      'Murray'
    )
  `

  const [account] = await sql`
    insert into accounts (
      user_id
    ) values (
      ${ user.user_id }
    )
  `

  return [user, account]
})
```

The above example is a bit confusing for someone who is new to postgres.js but I also believe that it does not work as intended.
From tests I have done, the return value of an `INSERT` query using the ```sql`insert into table ...` ``` function is an empty array (output straight from my terminal):
`Result(0) []`.

(Here I attach a screenshot of my output)

<img width="254" alt="image" src="https://github.com/porsager/postgres/assets/103976925/c7b0eef1-44b6-4779-974d-90246ef89c80">

I think the best way to remedy this is to use `RETURNING` in the `INSERT` query in function.

```js
const [user, account] = await sql.begin(async sql => {
  const [user] = await sql`
    insert into users (
      name
    ) values (
      'Murray'
    )
returning *
 `

  const [account] = await sql`
    insert into accounts (
      user_id
    ) values (
      ${ user.user_id }
    )
returning *
 `

  return [user, account]
})
```
[More info on RETURNING](https://stackoverflow.com/a/2944481/18999644)

This is a minor modification but debugging this has taken a couple of hours for me as I am slightly new to SQL syntax and to postgreSQL in general. I was trying to use the empty return from `sql.begin` but it turned out that the callback in `sql.begin` was the one returning the empty array even though the insert was successful.

I would love if this could be added to help new people who are probably going to check the readme for documentation. If it is needed may I submit a pull request too?

porsager:
Very nice description, and you are absolutely right! It's very confusing that it's coded as if there would be a resulting user from the insert when the `returning` clause is missing ! Sorry that got you sidetracked, a PR is very much welcome.

PR #662, opened 2023-08-20
Update README.md (Transactions added missing returning *)

MarisaCodes:
Check issue: https://github.com/porsager/postgres/issues/649

This is a minor modification but debugging this has taken a couple of hours for me as I am slightly new to SQL syntax and to postgreSQL in general. I was trying to use the empty return from sql.begin but it turned out that the callback in sql.begin was the one returning the empty array even though the insert was successful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,6 +579,7 @@ const [user, account] = await sql.begin(async sql => {
     ) values (
       'Murray'
     )
+  returning *
   `
 
   const [account] = await sql`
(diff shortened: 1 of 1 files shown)


porsager:
Thanks a lot @MarisaCodes , and sorry that caused so much trouble for you 😞

MarisaCodes:
> Thanks a lot @MarisaCodes , and sorry that caused so much trouble for you 😞

No it didn't it was quite a simple oversight on my part too since I was new to postgres. Thank you!!

PR #673, opened 2023-09-11
Fix connection on deno 1.36.3

JorritPosthuma:
Running the following code on deno 1.36.3 using postgres 3.3.5:

```javascript

const sql = postgres();

sql`SELECT 1`.then((result) => {
  console.log(result);
});

export default sql;
```

Gave the following error:

```
error: Uncaught (in promise) TypeError: Class constructor Socket cannot be invoked without 'new'
    at createSocket (./node_modules/.deno/postgres@3.3.5/node_modules/postgres/src/connection.js:131:15)
    at Timeout.connect [as _onTimeout] (./node_modules/.deno/postgres@3.3.5/node_modules/postgres/src/connection.js:328:31)
    at cb (ext:deno_node/internal/timers.mjs:63:31)
    at Object.action (ext:deno_web/02_timers.js:153:11)
    at handleTimerMacrotask (ext:deno_web/02_timers.js:67:10)
    at eventLoopTick (ext:core/01_core.js:189:21)
```

Simple fix is as proposed. Also tested if it still works in node v20.5.1, which it does :)

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -129,7 +129,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     try {
       x = options.socket
         ? (await Promise.resolve(options.socket(options)))
-        : net.Socket()
+        : new net.Socket()
     } catch (e) {
       error(e)
       return


porsager:
Ah, I see - so deno has started consuming Postgres.js directly since their npm support launch? (curious why the won't allow it since Node.js does).

Even so, no reason we shouldn't make the change since node seems to prefer `new` as well.

porsager:
Thank you

hastebrot:
Great work!

I've replaced

```
/// <reference types="https://deno.land/x/postgresjs@v3.3.5/types/index.d.ts" />
import postgres from "https://deno.land/x/postgresjs@v3.3.5/mod.js";
```

with

```
import postgres from "npm:postgres@3.4.0"
```

and it works like a charm with full typing support.

Interestingly, my tests now run faster; could be because it now uses Deno's Node polyfills and they maybe implemented some part of it in Rust.

porsager:
That was a great way to get the setup without my custom polyfills tested 😊 

Did you not find speedups anyway since you crossed it? I'd be curious to remove my hacks in place of Denos, but only if no slowdowns are incurred.

PR #694, opened 2023-10-13
Update Cloudflare createHash polyfill to support md5 and hex encoding

a-robinson:
Since the md5 method in cf/src/connection.js expects to be able to call crypto.createHash('md5').update(x).digest('hex')

This was causing md5 password auth to hang when used from a Cloudflare worker, but now I've confirmed md5 password auth works.


porsager:
Thanks a lot @a-robinson !

PR #707, opened 2023-10-21
Add common parameter names to ConnectionParameters TS type

### LINKED ISSUE #645: How to set DATESTYLE on new connections
Hi,

I'm looking for a way to automatically set DATESTYLE on all new connections opened by the node pool, to facilitate parsing timestamp values into temporal instants, but I'm not finding anything that could do that.

node-postgres has the [`connect`](https://node-postgres.com/apis/pool#connect) event on it's node pool that allows executing sql on all new connections before they are used, is there anything similar, or would you be open to adding that sort of functionality?

mkubliniak:
fixes #645

- added reference to [Client Connection Defaults PostgreSQL documentation](https://www.postgresql.org/docs/current/runtime-config-client.html) to the readme
- added common (often used from my experience) parameter names directly to the `ConnectionParameters` type
- changed value type of parameter from just `string` to `string | number | boolean` - number and boolean works too

### DIFF
--- a/README.md
+++ b/README.md
@@ -975,7 +975,7 @@ const sql = postgres('postgres://username:password@host:port/database', {
   },
   connection           : {
     application_name   : 'postgres.js', // Default application_name
-    ...                                 // Other connection parameters
+    ...                                 // Other connection parameters, see https://www.postgresql.org/docs/current/runtime-config-client.html
   },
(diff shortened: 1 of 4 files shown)


porsager:
Cool - thanks!

PR #736, opened 2023-11-22
Remove FLUSH from CF integration

tmthecoder:
The intermediate use of `FLUSH` in the current implementation for Cloudflare workers causes incompatibilities with Hyperdrive's caching service.

In order to support caching, this PR aims to remove the intermediate use of `FLUSH` when executing extended Postgres queries favoring sending the query all-together. This allows Hyperdrive's caching service to cache the full queries sent and serve results from cache for extended queries, resulting in less round-trips overall.

I believe all existing tests should pass (I ported the connection.js changes onto the deno implementation and ran the tests there). I also ran local tests to ensure compatibility.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -4,7 +4,7 @@ import crypto from 'crypto'
 import Stream from 'stream'
 import { performance } from 'perf_hooks'
 
-import { stringify, handleValue, arrayParser, arraySerializer } from './types.js'
+import { stringify, handleValue, arrayParser, arraySerializer, typesMapped } from './types.js'
 import { Errors } from './errors.js'
 import Result from './result.js'
 import Queue from './queue.js'
@@ -183,13 +183,13 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
 
     return q.options.simple
       ? b().Q().str(q.statement.string + b.N).end()
-      : q.describeFirst
+      : q.describeFirst 
         ? Buffer.concat([describe(q), Flush])
-        : q.prepare
-          ? q.prepared
-            ? prepared(q)
-            : Buffer.concat([describe(q), prepared(q)])
-          : unnamed(q)
+        : q.prepare 
+          ? q.prepared 
+            ? prepared(q) 
+            : Buffer.concat([describe(q), prepared(q)]) 
+          : unnamed(q);
   }
 
   function describe(q) {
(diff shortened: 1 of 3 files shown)


tmthecoder:
Ah alright, thanks for letting me know! Would you mind letting me know what the perf testing looked like? I'd love to reproduce it on my end to see the results if possible. 

I'll look into integrating this into the main src directory as well then

porsager:
Yeah, check out https://github.com/porsager/postgres-benchmarks

tmthecoder:
@porsager Should be updated for all tests and in the main source directory. The changes made essentially check if existing known type mappings exist and cut the round trip used by the `FLUSH` query. If the types aren't known by us, the `FLUSH` is sent and that round trip occurs. I'm unable to get the benchmarks running locally currently, but will try to soon to see how it looks in terms of performance, I'm curious to know what it looked like when you ran them

tmthecoder:
@porsager Benchmarks look like the fluctuate between this implementation and the existing implementation, so I don't see a visible constant difference on my end

tmthecoder:
In that case then, where no inference is needed, would there still be a need to receive information from the `ParameterDescription` response? If not, then this should still apply so we don't wait on the response from server when not needed

tmthecoder:
@porsager Ah my bad, I misunderstood then. I'd still be in favor of merging this PR for now as it seems like the JS inference will take a while to deprecate completely (a new major version release it seems), and if any other inference is thought out in the future this would still apply if or when that is implemented.

elithrar:
@porsager is this something you'd still want to merge + tag a release on? We'd love to get Postgres.js working w/ Hyperdrive + Workers again!

porsager:
Definitely! I've just been too busy to look at it, but today and tomorrow I'll be diving into these things again!

john-griffin:
Would also like to use Hyperdrive caching, any idea where the commit that closed this issue ended up? When clicking through it says "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

porsager:
Hi John. The PR didn't solve the issue, it's a bit deeper than that, and requires ignoring the describe part of the protocol which will change the behavior of Postgres.js. I'm talking with CloudFlare to see if we can find a good solution.

john-griffin:
@porsager thanks for clarifying 🙏

Issue #641, opened 2023-07-15
ECONNREFUSED inconsistency


mattbishop:
`sql` client instances handle ECONNREFUSED errors inconsistently. This error is thrown when the database is unreachable. In the steps below, I am in control of the postgres process. I would think the same steps work with network access as well.

### The Good

1. Launch Postgres.
2. Launch Node app and create client instance: `const sql = postgres(opts)`.
3. Shut down the Postgres process.
4. Client uses `sql` instance to run queries. These fail, as expected.
5. Relaunch Postgres.
6. Client uses `sql` instance to run queries. They succeed.

### The Not-So-Good

1. Shut down Postgres, or ensure it is not running.
2. Launch Node app and create client instance: `const sql = postgres(opts)`. Node throws this error:
```
Error: connect ECONNREFUSED 127.0.0.1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1592:16)
    at cachedError (file:///Users/matt/Projects/pg-test/node_modules/postgres/src/query.js:171:23)
    at new Query (file:///Users/matt/Projects/pg-test/node_modules/postgres/src/query.js:36:24)
    at sql (file:///Users/matt/Projects/pg-test/node_modules/postgres/src/index.js:111:11)
    at Function.listen (file:///Users/matt/Projects/pg-test/node_modules/postgres/src/index.js:179:35)
```
3. Client uses `sql` instance to run queries. These fail, as expected.
5. Launch Postgres.
6. Client uses `sql` instance to run queries. These fail. This is unlike the first set of steps, where the connection is re-established and queries proceed as normal.

So should they behave the same? If this is expected behaviour, what is the right pattern to use to keep the `sql` instance valid?

porsager:
Very nice description ! That is definitely not intended, I'll see if I can find some time to repro and look into it.

mattbishop:
I test this issue with every release, and it is still present in v3.4.2

micalevisk:
@mattbishop does v3.4.7 fixed it?

PR #738, opened 2023-11-24
Patch: Connection stuck after a while

wackfx:
Hey @porsager

I noticed that, after a while, I had random "WRITE_TIMEOUT" // "CONNECT_TIMEOUT" happening and my system became unresposive.

It appears that the socket, once closed, doesn't correctly reconnect. By recreating the socket, it looks like it solved the issue. Today, this behavior is done only on "secured" but I believe we could have it for all connections.

I believe it's ok to do so as 1) it shouldn't happen often 2) socket is closed and will eventually be garbage collected 3) it's already done for secured connections

## Reproducing & checks
Context: I'm using Bun
I started with Postgres down - and started postgres after the first attempt to try to reproduce #641

**⚠️ It looks like this behavior doesn't happen on Node** - when using node with or without my patch, the socket correctly reconnects. I believe it may be solved by some Node shenanigan and impacting Bun (or even Worker?) only.

```javascript

const sql = postgres(`${process.env['DATABASE_URL']}`, {
    max: 1, # trigger the issue sooner
    max_lifetime: 1, # trigger the issue sooner by closing the connection really quickly
})

setInterval(async () => {
    console.log((await sql`SELECT * FROM pg_am`)[0])
}, 3000)
```

### Before
After the first query - connection is closed (because of `max_lifetime: 1`) and doesn't reconnect
image

### After
It reconnects well.
image

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -429,10 +429,9 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     lifeTimer.cancel()
     connectTimer.cancel()
 
-    if (socket.encrypted) {
+    if (socket.encrypted)
       socket.removeAllListeners()
-      socket = null
-    }
+    socket = null
 
     if (initial)
       return reconnect()
--- a/src/connection.js
+++ b/src/connection.js
@@ -429,8 +429,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     lifeTimer.cancel()
     connectTimer.cancel()
 
-    if (socket.encrypted)
-      socket.removeAllListeners()
+    socket.removeAllListeners()
     socket = null
 
     if (initial)


dimonnwc3:
@wackfx thanks for contribution. I've experienced the same, than connection in bun times out and then postgres is not available anymore, with the following error: `write CONNECT_TIMEOUT postgres:5432`

before it gets merged, a workaround is to use a "secured" connection?

wackfx:
That's correct.

Usng a secured connection (so with SSL connection, or give a custom encrypted socket at connection time) would produce the same behaviour as this PR.
-------- Message d'origine --------

porsager:
Very nice debugging, and thanks for the proposed fix @wackfx, but I actually think this should be fixed on Buns end. Making workarounds for Bun bugs in all libraries is just a useless work multiplier. If we were using native Bun apis, it would be a completely different matter, but this is Bun wanting to mimick Node.js APIs, so I'm certain they'd also want to have a bug like this fixed. - Did you make an issue with Bun, and a minimal reproducible case using only `net` (tcp)? It might be an even simpler fix on their end.

porsager:
Even so, you are also right, there shouldn't be any actual performance concerns or issues recreating the socket for non ssl connections too, so let's take your simplification 😉

wackfx:
hey @porsager - you're very right about the fact that Bun might wanna take a look into this for a better Node compat, tbh I opened this before trying to reproduce with Node.JS.

dimonnwc3:
@porsager any plans on releasing this to npm?

Th1nkK1D:
Amazing findings! I ran into the same issue. While waiting for a release, I use the following command to temporarily patch `connection.js` from this PR.

```sh
curl https://raw.githubusercontent.com/porsager/postgres/6f20a4820c683b33e7670b606d8daf5670f4b973/src/connection.js > node_modules/postgres/src/connection.js
```

PS. I added it in my docker file after `bun install`, but you and also run it manually.

mattbishop:
Probably can remove the two individual `socket.removeListener()` calls a few lines before this change too.

WildEgo:
Can we publish this..? It's an high impact bug fix

hex2f:
any update on when this will make it into a release? its causing me a lot of headaches right now 😅

amjed-ali-k:
any update ❓

emilienbidet:
It cause me a lot of troubles too. Latest release is from November, may we can make a new one?

porsager:
A reminder at the right time ;) Just released v3.4.4

edwardchew97:
> A reminder at the right time ;) Just released v3.4.4

Deployed the latest version yesterday, 12 hours passed and so far it seems to resolve the issue in our case! 🙌 
image

porsager:
Fantastic! Mind sharing what you're running on? Bun?

porsager:
@edwardchew97 are you using bun?

kam-st:
Hello, so pulling directly from github to resolve cloudflare issue. But the issue of queries getting stuck persists. The way I reproduce stuck queries is through repetitive refresh of page. The queries though authjs gets stuck and whole server gets jammed. 

Stack is NextJS | Drizzle | AuthJS (Drizzle running through postgresjs driver)

The reason I know this issue is with postgresjs is because if I use neon drivers. Then the issue does not persists and everything works flawlessly.

gregory-krieger:
This is the solution I implemented, just replace this :

```
import "dotenv/config";

import { drizzle } from "drizzle-orm/postgres-js";
import * as schema from "./schema";

const connectionString = process.env.DATABASE_URL;

// Disable prefetch as it is not supported for "Transaction" pool mode
export const client = postgres(connectionString!, {
  prepare: false,
  ssl: false,
});
export const db = drizzle(client, { schema });
```

with this :

```
import "dotenv/config";
import { PostgresJsDatabase, drizzle } from "drizzle-orm/postgres-js";
import * as schema from "./schema";

declare global {
  var database: PostgresJsDatabase<typeof schema> | undefined;
}

const connectionString = process.env.DATABASE_URL;

if (process.env.NODE_ENV === "production") {
  database = drizzle(
    postgres(connectionString!, {
      prepare: false,
    }),
    { schema },
  );
} else {
  if (!global.database)
    global.database = drizzle(
      postgres(connectionString!, {
        prepare: false,
      }),
      { schema },
    );
  database = global.database;
}
export const db = database;
```

PR #770, opened 2024-01-06
Add `simple()` type definition

### LINKED ISSUE #714: sql.simple() for DB initialization and typescript support
# Problem

I have a use-case where I have a DDL script to initialise the database. Something similar to below:

```sql
CREATE DATABASE dbName with owner dbuser ;

CREATE TABLE IF NOT EXISTS table (
  col1 VARCHAR(255) NOT NULL,
  col2 VARCHAR(255) NOT NULL,
);
```

I'm trying to use it as below:
```typescript
await (sql`${sql.unsafe(sqlScript)}` as any).simple();
```

which results in the following error:

```
Error creating database PostgresError: CREATE DATABASE cannot run inside a transaction block
```

EDIT: I realize this is not possible to do in postgresql since one cannot switch a database without re-initializing the connection (aka backend).

# Questions: 
- What is the recommended way to do this? I'd like to maintain the multiple statement SQL script independently.
- Can we get typescript support for `simple` method?

iby:
Related to #714.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -599,6 +599,7 @@ declare namespace postgres {
   type RowList<T extends readonly any[]> = T & Iterable<NonNullable<T[number]>> & ResultQueryMeta<T['length'], keyof T[number]>;
 
   interface PendingQueryModifiers<TRow extends readonly any[]> {
+    simple(): this;
     readable(): Promise<Readable>;
     writable(): Promise<Writable>;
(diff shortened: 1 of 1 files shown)


rmnoon:
Just ran into this myself! Almost opened another issue. cc @porsager

porsager:
Thanks :)

Issue #571, opened 2023-03-16
Specify CA certificate manually


baoshan:
With `deno-postgres`, I can [specify a `.crt`](https://deno.land/x/postgres@v0.17.0/docs/README.md?source=#about-invalid-and-custom-tls-certificates) file (e.g., `supabase.crt`), and `sslmode=require` works (tested just now).

Is this an option for `Postgres.js`?

ralfstx:
BTW, in Node.js you can use `{ ssl: { ca: [ fs.readFileSync('server-cert.pem') ] }}`, see https://nodejs.org/api/tls.html

oscar6echo:
> From the postgres lib point of view, is there a workaround ? even hacky !

Upon further [investigation ](https://github.com/denoland/deno/issues/28903#issuecomment-2809728945)this is independent of the posgres lib.

PR #771, opened 2024-01-07
deno: move from std node polyfills to `node:` specifier

Kyiro:
# Hello

I've been using postgres.js in a Deno project for a while but I just noticed how outdated the compatibility really is.

Before this gets accepted though, I'd also like to solve these errors when the connection string is invalid/can't connect. I have no idea if these happen on node but I'd assume not.
```ts
error: Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
      stack: { value: err.stack + query.origin.replace(/.*\n/, '\n'), enumerable: options.debug },
                                               ^
    at queryError (file:///C:/Projects/postgres/deno/src/connection.js:391:48)
    at errored (file:///C:/Projects/postgres/deno/src/connection.js:386:17)
    at Socket.error (file:///C:/Projects/postgres/deno/src/connection.js:378:5)
    at Socket.emit (ext:deno_node/_stream.mjs:1852:9)
    at emitErrorNT (ext:deno_node/_stream.mjs:1572:13)
    at emitErrorCloseNT (ext:deno_node/_stream.mjs:1544:7)
    at processTicksAndRejections (ext:deno_node/_next_tick.ts:33:15)
    at runNextTicks (ext:deno_node/_next_tick.ts:71:3)
    at eventLoopTick (ext:core/01_core.js:189:21)
```


mufaroxyz:
same 👍

Kyiro:
It'd honestly be good to get this merged and resolve the error issue at a later date because of those warnings, could we get this PR reviewed?

porsager:
I'm thinking a hard cut in v4 where we just remove the deno polyfill completely if possible?

mdluo:
> As Deno got npm compatibility a while back, I think it might be the right play but this could still get accepted for v3 to avoid headaches

Yes, the current version on [supabase/edge-runtime](https://github.com/supabase/edge-runtime) it's not just a warning, but  it triggers an error and become unusable:

```log
runtime has escaped from the event loop unexpectedly: event loop error: TypeError: internals.warnOnDeprecatedApi is not a function
    at get rid (ext:deno_io/12_io.js:277:15)
    at createWritableStdioStream (https://deno.land/std@0.132.0/node/_process/streams.mjs:29:21)
    at https://deno.land/std@0.132.0/node/_process/streams.mjs:68:38
```

While switching to use this PR the error is gone.

lukeed (approved):

ardabeyazoglu:
This is reproduced on both nodejs and deno. Is there any timeline to merge ?

ardabeyazoglu:
Note that removing tls polyfill cause ssl connections fail to a database with self signed certificate. This breaks cloud services like Supabase as well. The reason for that is the incompatibility of tls socket options in different runtimes. In nodejs, you set `ca: ["base64 certificate content"]` in postgres connections options.ssl to mitigate this. However, this didn't work in my Deno setup. Using the polyfill and setting `caCerts: ["base64 certificate content"]` the same way fixed issue.

Sleepful:
the `ca` and `caCerts` difference can be noted in this other comment thread: https://github.com/porsager/postgres/issues/571#issuecomment-1476982783

Issue #757, opened 2023-12-13
CONNECTION_CLOSED on subscribe()


dvv:
I use standard
```ts
const { unsubscribe: _unsubscribe } = await sql.subscribe(
  "*",
  (row, info) => {
    console.log("SUB", row, info)
  },
  () => {
    console.log("SUB!")
  },
)
```

and it works well.

When postgres goes down I immediately get
```
Unexpected error during logical streaming - reconnecting Error: write CONNECTION_CLOSED ...
    at closed (https://deno.land/x/postgresjs@v3.4.2/src/connection.js:443:57)
    at https://deno.land/x/postgresjs@v3.4.2/polyfills.js:138:30
    at Array.forEach (<anonymous>)
    at call (https://deno.land/x/postgresjs@v3.4.2/polyfills.js:138:16)
    at closed (https://deno.land/x/postgresjs@v3.4.2/polyfills.js:127:5)
    at success (https://deno.land/x/postgresjs@v3.4.2/polyfills.js:109:7)
    at eventLoopTick (ext:core/01_core.js:189:11) {
  code: "CONNECTION_CLOSED",
  errno: "CONNECTION_CLOSED",
```

When postgres goes up then I get everything working except for the above subscription -- it never reconnects.

Louis-Tian:
Just got caught with the same error. And it's causing my database to bloat with the WAL piling up...

Louis-Tian:
I believe the key problem the "subscribe" creates a *temporary* replication slot on the database. As soon as the database shuts down, that original replication slot is gone forever. Postgres.js at the moment is not smart enough to recreating a new replication slot when the database is back online. 

@porsager What is really problematic at the moment is that there is no way for us to handle it ourselves in this situation. As the error is being completely swallowed by the stream here.
https://github.com/porsager/postgres/blob/6f20a4820c683b33e7670b606d8daf5670f4b973/src/subscribe.js#L99-L107

I think even just rethrow the error would be better. Because then I can catch that error using the global "UncaughtException" event. And even if I don't, the program would completely crash and restarted by the container runtime or process manager alike.

dvv:
@Louis-Tian right.
I also drafted https://github.com/porsager/postgres/compare/master...dvv:postgres:master to move "close" logic uplevel.

smartcrash:
I having this same issue. Have you found a work around for this?

X-43A:
Same issue. A solution would be appreciated. Thanks

hanszoons:
[...]

PR #785, opened 2024-01-24
add error handler to subscribe

Louis-Tian:
Related to https://github.com/porsager/postgres/issues/757.

Expose a error handler on the subscribe function so any error happened while streaming replica can be handled in user space.

### DIFF
--- a/src/subscribe.js
+++ b/src/subscribe.js
@@ -47,7 +47,7 @@ export default function Subscribe(postgres, options) {
 
   return subscribe
 
-  async function subscribe(event, fn, onsubscribe = noop) {
+  async function subscribe(event, fn, onsubscribe = noop, onerror = noop) {
     event = parseEvent(event)
 
     if (!connection)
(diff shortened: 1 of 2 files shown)


mattbishop:
Can you update the Typescript definition to include this please?

Louis-Tian:
@mattbishop done. Happy to add a test case for it as well. 

But I am not sure what would be best way to go about it. Specifically what's the best way to stub the stream in test to force an error?

porsager:
@Louis-Tian I have some other tests where I just terminate the backend. Perhaps that's a way forward?

Louis-Tian:
@porsager That's kind of what I am doing in the error handler anyway so that will work for me. Just think it would be better to let the user to decide to what do? But if there is other things that require a direct termination then I guess it make perfect sense to just leverage on that.

porsager:
This already seems to be merged, so I don't understand why the PR is still open.

PR #618, opened 2023-06-23
draft: Cloudflare works ! 🎉 

wackfx:
Hey there !

Amazing work with your lib. It's so awesome that I was really sad to see that it didn't work on cloudflare 😢 

I've spent some time to make it work on Cloudflare. Tested on my own project - it's working for both Cloudflare pages workers and regular workers.

This PR is **not** ready to merge - but since I've spent hours trying to deploy this library on Cloudflare based on your work - I guess you may want to take some pointers to what's missing from your implementation to successfully deploy it.

### Here's what changed:
- Rebased your current work on `master`
- polyfill: `Crypto.subtle.digest(MD5)` didn't work as md5 is not supported anymore. Added Joseph's Myers implementation to follow your pattern of 'no external libs'
- polyfill: Added dynamic import of `cloudflare:socket` to please bundlers
- transpile: Replacing `net.Socket()` with `async net.Socket()` to allow for dynamic import on CF code
- polyfill: `os` `fs` polyfills that doesn't respect the original contract - but since we're on cloudflare there is not much more we can do
- polyfill: `process.hrtime`
- transpile: Buffer being undefined on cloudflare - added import on build:cf
- transpile: added `node:` prefixes

### How to use ?

Simply `import postgres from 'postgres'` and instantiate it like you would do in a regular node env.

I use this module with Qwik and the adapter cloudflare-pages.
You'll need to tell bundler (through vite.config.ts on Qwik) to resolve `workerd` and set the node library (but compatible with cf workers) as external like this:
image

That's it. Feel free to take whatever you want and close this PR, I won't take it personally, I know it's pretty hacky ;)
Thanks again for your initial work.

PR #819, opened 2024-02-28
chore: update export conditions

Hebilicious:
Hello there !
The official runtime export key for cloudflare is `workerd` (not worker).
I believe many apps out there might already be relying on `worker` so I propose to add it alongside it.

Reference : 
- https://developers.cloudflare.com/workers/wrangler/bundling/#conditional-exports

- https://runtime-keys.proposal.wintercg.org/#workerd

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "types": "./types/index.d.ts",
     "bun": "./src/index.js",
     "worker": "./cf/src/index.js",
+    "workerd": "./cf/src/index.js",
     "import": "./src/index.js",
     "default": "./cjs/src/index.js"
   },


porsager:
Read from here on down for why it was changed from workerd to worker originally. https://github.com/porsager/postgres/pull/618#issuecomment-1693173590

Would be great to get a confirmation from cloudflare if they now think `workerd` is the way to go before merging.

elithrar:
@porsager - `workerd` is the correct export. Also see https://runtime-keys.proposal.wintercg.org/#workerd

porsager:
I'll merge this and remove the worker export for now, as it appears the cases where the worker export is requested actually needs the default esm export.

PR #829, opened 2024-03-19
Improve `onnotice` to actually not log anything when its config value is `false`

hughcrt:
This options can also accept false as a value, not only a function


hughcrt:
Hi @porsager, could you please merge this PR?

porsager:
A noop function `() => {}` is fine imo to ignore output

PR #863, opened 2024-05-09
Update README.md -- add `sql.array` method

Marviel:


porsager:
sql.array is deprecated

Marviel:
@porsager thanks for the update -- how would you recommend we use multidimensional arrays?

porsager:
Are you having issues with that?

PR #870, opened 2024-05-16
Add correct onnotice silence value in README.md

hughcrt:
I previously made a [PR](https://github.com/porsager/postgres/pull/829) to fix the types of `onnotice` by adding false as a correct value. This was because the README said so. 
@porsager you said that a noop function was fine, so I've updated the Readme.


hughcrt:
@porsager, could you please merge this PR?

Issue #881, opened 2024-05-28
Feature request: dynamic username support for credential rotation


izeau:
Greetings!

I have noticed that the `password` option can be set to a function, allowing for a seamless way to reconnect when credentials are rotated. In my current workflow, we also rotate usernames. For example, we use `user_1` and `user_2`, but only one of them is allowed to log in at a given time (let's say it's `user_1`).

When we need to rotate credentials, we generate a random password for `user_2` and enable it for login. We then update the secrets with `username = user_2` and the new password, notifying applications of the credential change. After a brief grace period, `user_1` is disabled until the next rotation.

My issue is that while the `password` can be dynamically set, the `username` cannot. Is there a workaround for this, or could support for a dynamic `username` option be implemented? This feature would greatly enhance the flexibility and security of our credential management process.

Thank you for considering this request.

porsager:
Yeah, that's a good idea :) feel free to do a pr mimicking the password behavior ;)

PR #882, opened 2024-05-28
allow user & database to be provided dynamically

izeau:
Solves #881 

Please let me know if I missed something :) Thanks!


porsager:
quick first look, but you can leave out the changes in all the folders that are built on release ;) Makes the PR less noisy

izeau:
Sure thing!

PR #901, opened 2024-06-24
fix(types): use correct default export

### LINKED ISSUE #862: Incorrect type definitions causing confusion when trying to catch PostgresError
I noticed that when I want to catch an error and identify it using `instanceof` I got an error `TypeError: Right-hand side of 'instanceof' is not an object`. Which was caused by following code:

```ts
import { PostgresError } from 'postgres';

try {
  // await query
} catch (e) {
  if (e instanceof PostgresError) {
    // do something
  }
}
```

To fix that I had to use default import:

```ts
import Postgres from 'postgres';

try {
  // await query
} catch (e) {
  if (e instanceof Postgres.PostgresError) {
    // do something
  }
}
```

This is confusing because typescript was not able to catch the failing example as something wrong and I was able to find this error only in runtime.

plushdohn:
This PR fixes #862 by ensuring classes defined under the `postgres` namespace aren't also available as a named export, but only under the default `postgres.` export.

I've tested the change by patching the module in my projects, but I'm not sure if under some specific Typescript scenarios the behaviour could be incorrect.

---

Trying to import `PostgresError` as a named export:

Screenshot 2024-06-24 at 12 30 01

---

And here's the correct usage using the default export:

Screenshot 2024-06-24 at 12 31 14

### DIFF
--- a/deno/types/index.d.ts
+++ b/deno/types/index.d.ts
@@ -728,4 +728,4 @@ declare namespace postgres {
   }
 }
 
-export = postgres;
+export default postgres;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -726,4 +726,4 @@ declare namespace postgres {
   }
 }
 
-export = postgres;
+export default postgres;


blechatellier:
Would like this one merged please 👍

porsager:
Needs some more thumbs up from other typescript users to confirm this doesn't break anything? (i don't use typescript so can't tell)

Issue #807, opened 2024-02-15
postgres.js needs something like slonik's sql.join() for building dynamic queries


DanielFGray:
postgres.js gives us parameterized tagged templates and the ability to nest sql partials, and this is fantastic, but I think it is missing a piece of the puddle:

I ~~want~~ *need* to be able concat partials together (that may contain query parameters), usually with some delimiter (eg whitespace, or a comma, or maybe an sql operator)

slonik provides something like:
```js
sql.join = (a: PendingQuery[] | Helper[], p: PendingQuery): PendingQuery {...}
```
and I think it would be amazing if this were a first-class feature in postgres.js

this should allow things like
```js
let dynamicFilters = [
  sql`somecol = ${somevalue}`,
  sql`col2 = ${v2}`
];
const dynamicColumns = [sql('somecol'), sql('col2')]
await sql`
  select ${sql.join(dynamicColumns, sql`, `)} from t
  where ${sql.join(dynamicFilters, sql` and `)};
`;
```

porsager:
Yeah, I think it makes sense to add, but I've just been doing this when needed

```js
xs.flatMap((x, i) => i ? [sql`, `, x] : x)
```

You can fairly quickly make a function to use instead if you feel like it - eg 😋

```js
sql.join = (xs, joiner) => xs.flatMap((x, i) => i ? [joiner, x] : x)
```

I'll keep the issue open for now.

DanielFGray:
> `sql.join = (xs, joiner) => xs.flatMap((x, i) => i ? [joiner, x] : x)`

oh interesting! i would never have come up with that on my own

do you have any reservations about adding this as a core feature?

DanielFGray:
and if no reservations, would you accept a PR adding this as a core feature?

davepar:
It would be great to at least add the work-around to the docs. It wasn't obvious to me how to join sql fragments with a comma. Glad I found this.

DanielFGray:
i'm working on a PR

anyone want to bike-shed over `join` vs `concat` to avoid confusion with sql joins?

Destreyf:
> Yeah, I think it makes sense to add, but I've just been doing this when needed
> 
> ```js
> xs.flatMap((x, i) => i ? [sql`, `, x] : x)
> ```
> 
> You can fairly quickly make a function to use instead if you feel like it - eg 😋
> 
> ```js
> sql.join = (xs, joiner) => xs.flatMap((x, i) => i ? [joiner, x] : x)
[...]

PR #905, opened 2024-06-27
feat: sql.join

DanielFGray:
this is a draft fix for #807

I couldn't get the tests to run so they probably need some work

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,11 @@ select * from users
 select * from users where user_id = $1
 ```
 
+Multiple fragments can be joined together with `sql.join` to create more complex dynamic queries:
+```js
+const columns = ['id', 'name', 'age'].map(name => sql(name))
+sql`select ${sql.join(', ', columns)} from users where
+
 ### SQL functions
 Using keywords or calling functions dynamically is also possible by using ``` sql`` ``` fragments.
 ```js
--- a/src/index.js
+++ b/src/index.js
@@ -98,6 +98,7 @@ function Postgres(a, b) {
       notify,
       array,
       json,
+      join,
       file
     })
 
@@ -316,6 +317,10 @@ function Postgres(a, b) {
 
   function json(x) {
     return new Parameter(x, 3802)
+  }
+
+  function join(sep, xs) {
+    return xs.flatMap((x, i) => i ? [sep, x] : x)
   }
 
   function array(x, type) {
--- a/tests/index.js
+++ b/tests/index.js
@@ -2469,6 +2469,14 @@ t('Supports arrays of fragments', async() => {
   ]
 })
 
+t("Joins fragments with a separator", () => {
+  const fs = [sql`a = ${1}`, sql`b = ${"test"}`]
+  return [
+    sql`select * from t where ${ sql.join(sql` and `, fs) }; `.describe().string,
+    'select * from t where a = $1 and b = $2'
+  ]
+});
+
 t('Does not try rollback when commit errors', async() => {
   let notice = null
   const sql = postgres({ ...options, onnotice: x => notice = x })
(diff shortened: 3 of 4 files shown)


DanielFGray:
if you have a weird workaround you can try this fix in your own projects

```sh
npm i github:danielfgray/postgres.js
```
@Destreyf @davepar

src/index.js
```diff
@@ -318,6 +319,10 @@ function Postgres(a, b) {
     return new Parameter(x, 3802)
   }
 
+  function join(sep, xs) {
+    return xs.flatMap((x, i) => i ? [sep, x] : x)
```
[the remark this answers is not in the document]
DanielFGray: [eba447a](https://github.com/porsager/postgres/pull/905/commits/eba447a13c9ccac34ebebf73244ee38f7e3ec917) should fix this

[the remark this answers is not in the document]
DanielFGray: ok latest commit is I think what you were actually suggesting

DanielFGray:
I also really want to enable currying to allow things like
```js
const and = sql.join(sql` and `)
and([sql`...`, ...fragments])
```
but this would also need further bike-shedding (does it need an array? should we slurp all params to `join`?)

src/index.js
```diff
@@ -318,6 +319,10 @@ function Postgres(a, b) {
     return new Parameter(x, 3802)
   }
 
+  function join(sep, xs) {
+    return xs.flatMap((x, i) => i ? Array.isArray(x) ? [sep, ...x] : [sep, x])
```
Destreyf, line 323: Is this missing the second ternery response?  right now it reads as you get either get the spread, or non spread version with the separator applied, but in the case of the "first" value, we would normally just want x by itself right?

```js
return xs.flatMap((x, i) => i ? (Array.isArray(x) ? [sep, ...x] : [sep, x]) : x )
```
added parathesis to show groupings, they're not needed for the expression to evaluate correctly.

DanielFGray: oof
i really wish i could get these tests running

README.md
```diff
@@ -342,6 +342,11 @@ select * from users
 select * from users where user_id = $1
 ```
 
+Multiple fragments can be joined together with `sql.join` to create more complex dynamic queries:
+```js
+const columns = ['id', 'name', 'age'].map(name => sql(name))
+sql`select ${sql.join(', ', columns)} from users where
```
davepar, line 348: I think this example line should be:

    sql`select ${sql.join(sql`, `, columns)} from users where

Doesn't the separator need to be a sql clause?

phosmium:
neat feature, hope this makes it in +rep

dhardtke:
I agree though that it should not be called "join" to avoid confusion with SQL's JOIN statements. In my opinion, "sql.concat" is a better choice.

maumercado:
```sql.concat > sql.join``` solid addition tho' 👍

DanielFGray:
> I agree though that it should not be called "join" to avoid confusion with SQL's JOIN statements. In my opinion, "sql.concat" is a better choice.

I've gone back and forth on this in my head, and while there's definitely potential for ambiguity, `join` here is behaving the same as it does on arrays

It would be even more surprising behavior to me to have something called `concat` behaving as `Array#join`, and if a user is confused on what `sql.join` does the docs would explain. I also think actually seeing the code in context would pretty well indicate it has nothing to do with the SQL clause

So all this is to say, semantically, I think join is the best name for this feature

jantoine1:
What's required to get this in? I also struggled discovering how to do this expecting the library to provide a mechanism.

DanielFGray:
> What's required to get this in? I also struggled discovering how to do this expecting the library to provide a mechanism.

getting the tests to pass is probably the main blocker

DanielFGray:
I consider this PR a success then, even if my implementation/API aren't used

DanielFGray:
> May I ask why you closed your PR?

@dhardtke the work in the PR itself is pretty trivial, and I don't currently have the time/energy to work on refactoring it, so I'll step aside rather than impede progress.

PR #912, opened 2024-07-07
Adding PGAPPNAME Environment Variable

gimse:
Making it possible to set the application_name with the PGAPPNAME environment variable. This is a standard set by the Postgres team (https://www.postgresql.org/docs/current/libpq-envars.html). 

It is a nice to have feature, and makes it easier to transition from the pg package 🔢 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 # Changelog
 
-## v3.3.0 - 9 July 2024
-- Adding support for the PGAPPNAME environment variable
-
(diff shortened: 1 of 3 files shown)


porsager:
Ah - this is great - thanks.. If you just remove the changelog line, I'll merge and add it once I cut a release.

gimse:
> Ah - this is great - thanks.. If you just remove the changelog line, I'll merge and add it once I cut a release.

Done :)

Issue #923, opened 2024-08-02
TypeError: Cannot read properties of undefined (reading 'replace')


alexgleason:
I'm getting some kind of error, but the error handler itself has an error, so I'm not sure what the error is. :smiley: 

```
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'replace')
      stack: { value: err.stack + query.origin.replace(/.*\n/, '\n'), enumerable: options.debug },
                                               ^
    at queryError (https://deno.land/x/postgresjs@v3.4.4/src/connection.js:392:48)
    at errored (https://deno.land/x/postgresjs@v3.4.4/src/connection.js:387:17)
    at error (https://deno.land/x/postgresjs@v3.4.4/src/connection.js:379:5)
    at https://deno.land/x/postgresjs@v3.4.4/polyfills.js:138:30
    at Array.forEach (<anonymous>)
    at call (https://deno.land/x/postgresjs@v3.4.4/polyfills.js:138:16)
    at error (https://deno.land/x/postgresjs@v3.4.4/polyfills.js:131:5)
    at eventLoopTick (ext:core/01_core.js:168:7)
```

alexgleason:
So I edited the code manually and logged it. `err` is:

```
InvalidData: invalid peer certificate: UnknownIssuer
    at async TlsConn.read (ext:deno_net/01_net.js:150:15)
    at async success (https://raw.githubusercontent.com/alexgleason/postgres.js/51ac395234089d5b7806c471830a1a472958fb06/deno/polyfills.js:97:56) {
  name: "InvalidData"
}
```

And the value of `query` is... `true` ??? Bro, this codebase needs TypeScript.

marbemac:
Running into this as well

patrickReiis:
I'm also having this problem

tilemanrenovations:
How long has this been happening?

dhardtke:
We are also seeing this error in our projects:
```
TypeError: Cannot read properties of undefined (reading 'replace')
    at queryError (/usr/src/app/node_modules/postgres/cjs/src/connection.js:389:48)
    at errored (/usr/src/app/node_modules/postgres/cjs/src/connection.js:383:14)
    at TLSSocket.data (/usr/src/app/node_modules/postgres/cjs/src/connection.js:318:9)
    at TLSSocket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
[...]

Issue #778, opened 2024-01-15
Deno - error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'replace')


bidipeppercrap:
![image](https://github.com/kysely-org/kysely-postgres-js/assets/30064028/df1a9d1a-95b2-4296-bb4f-bde76e42f2c8)
![image](https://github.com/kysely-org/kysely-postgres-js/assets/30064028/3422f25e-f2b1-46f8-a7a2-60af2a9114de)

## Spec

Runtime: **Deno**
Database Provider: **Neon**

### Reproduce:

I run into this error when I try to migrate.

## Code

deno.json
```json
{
    "imports": {
        "db/": "./src/db/",
        "kysely": "npm:kysely@^0.27.2",
        "kysely-postgres-js": "npm:kysely-postgres-js@^2.0.0",
        "postgres": "https://deno.land/x/postgresjs@v3.4.3/mod.js"
    },
    "tasks": {
        "migrate": "deno run -A ./src/db/migrator.ts"
    }
}
```

migrator:
```typescript
import * as path from "node:path";
import { fileURLToPath } from "node:url";
import { promises as fs } from "node:fs";
import { Migrator, FileMigrationProvider } from "kysely";
import { db } from "db/database.ts";

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);

async function migrateToLatest() {
    const migrator = new Migrator({
        db,
        provider: new FileMigrationProvider({
            fs,
            path,
            migrationFolder: path.join(__dirname, "./migrations")
        })
    });

    const { error, results } = await migrator.migrateToLatest();

    results?.forEach((it) => {
        if (it.status === "Success") {
            console.log(`migration "${it.migrationName}" was executed successfully`);
        } else if (it.status === "Error") {
            console.error(`failed to execute migration "${it.migrationName}"`);
        }
    });

    if (error) {
        console.error("failed to migrate");
        console.error(error);
        Deno.exit(1);
    }

    await db.destroy();
}

migrateToLatest();
```

db instance:
```typescript
import { Database } from "db/types/index.ts";
import postgres from "postgres";
import { Kysely } from "kysely";
import { PostgresJSDialect } from "kysely-postgres-js";
import { load } from "https://deno.land/std@0.212.0/dotenv/mod.ts";

const env = await load();

const dialect = new PostgresJSDialect({
    postgres: postgres({
        database: env["PGDATABASE"]!,
[...]

PR #944, opened 2024-09-16
Fix ambiguous error on startup

alexgleason:
Fixes https://github.com/porsager/postgres/issues/923

Before & after:

Screenshot from 2024-09-15 18-58-17

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -385,6 +385,8 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function queryError(query, err) {
+    if (typeof query !== 'object') throw err
+
     'query' in err || 'parameters' in err || Object.defineProperties(err, {
       stack: { value: err.stack + query.origin.replace(/.*\n/, '\n'), enumerable: options.debug },
       query: { value: query.string, enumerable: options.debug },
--- a/src/connection.js
+++ b/src/connection.js
@@ -385,7 +385,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function queryError(query, err) {
-    if (typeof query !== 'object') throw err
+    if (!query || typeof query !== 'object') throw err
 
     'query' in err || 'parameters' in err || Object.defineProperties(err, {
       stack: { value: err.stack + query.origin.replace(/.*\n/, '\n'), enumerable: options.debug },


patrickReiis (approved):
LGTM

martinwepner-otto:
Would be nice if this could be merged and released soon. In my team multiple developers ran into this when trying to connect with wrong credentials and the "cannot read property "replace" of undefined" was not very helpful.

fin-dogedev:
This runtime type error causes the application to terminate if the connection to the database fails unexpectedly. As a result, the reconnection mechanism cannot take effect.

coder-se:
Is there anything blocking this from being merged? Would be really good to get this into a release. The [issue](https://github.com/porsager/postgres/issues/778) has been around since beginning of this year causing applications to terminate like @fin-dogedev says. 

Any thoughts @patrickReiis , @porsager ?

chanon:
Please merge this.
Is this project still being maintained?

porsager:
also changes should only be made to the src, the rest of the files are auto generated

chanon:
FYI it also happens in the Node.js version I am using  `v20.13.1`

Also it is an uncaught exception that kills the process, so not sure how to write a test for it.

wataruoguchi:
During development, I had the same issue as https://github.com/porsager/postgres/issues/778. I can reproduce the issue with [my repository](https://github.com/wataruoguchi/poc-user-cf-hono-remix/pull/2).

- Prod runtime: Cloudflare Workers
- Dev runtime: Node (v22.3.0)
- Framework: Remix
- Database: Supabase
- Query Builder: Kysely

I confirmed the change in the `src/connection.js` (Node.js) in this PR has solved the issue.

I support this PR because the `queryError` function expects `query` to be an object. If the `query` is NOT an object, the `queryError` function fails anyway. I hope this PR gets merged.

**Side note:** During development, we need to use NodeJS's module by [adding an alias](https://github.com/remix-run/remix/issues/9245#issuecomment-2179517678).

alexgleason:
> also changes should only be made to the src, the rest of the files are auto generated

I updated to only include changes to src.

alexgleason:
I do not know how to reproduce an SSL connection error in a test. This issue is apparently also occurring in Node https://github.com/porsager/postgres/issues/923#issuecomment-2462024110

coder-se:
A note on this PR, we're running a fork of this in production and are still getting the ` Cannot read properties of undefined (reading 'replace')` error. So I think there might be other scenarios where this can happen. Likely it's `.origin` that's missing sometimes even if `query` is an object.

Still getting this error,

```
stack: { value: err.stack + query.origin.replace(/.*\n/, '\n'), enumerable: options.debug },
                                               ^
TypeError: Cannot read properties of undefined (reading 'replace')
    at queryError (file:///Users/alig/Projects/Validio/app/node_modules/postgres/src/connection.js:391:48)
    at errored (file:///Users/alig/Projects/Validio/app/node_modules/postgres/src/connection.js:383:14)
    at Socket.data (file:///Users/alig/Projects/Validio/app/node_modules/postgres/src/connection.js:318:9)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:392:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:189:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```

Maybe it's better to change the code so that origin is only appended if it's actually defined?

alexgleason:
@porsager This is a simple change that is 1 line of code and makes intuitive sense because it is just a guard clause. Will you please merge it?

porsager:
I'll take a look at it tonight @alexgleason !

coder-se:
We have been running a fork of this PR in production for awhile now, and there seems to be another case where query is actually is an object. But it doesn't have the `origin` property causing the same issue to be thrown. I think it would be good to adapt the code so that origin is only appended if the property exist.

alexgleason:
@coder-se I've never encountered that, but since it does seem possible for `query` to be null (eg here: https://github.com/porsager/postgres/blob/6ec85a432b17661ccacbdf7f765c651e88969d36/src/connection.js#L105) I added a check for it.

termermc:
This is causing issues for me as well (on Node and Deno). Thanks @alexgleason for your work on this, I hope it can be merged in soon.

danieldietzler:
Hey @porsager, is anything blocking this PR? We're using it in https://github.com/immich-app/immich and it'd be helpful if users can get proper error messages :) Thanks!

bo0tzz:
If it helps, this ^ linked pr shows the result of some testing of this patch.

theoludwig:
@porsager Could you please clarify what is blocking this PR to be merged and released?

porsager:
My time was the thing blocking it. I felt the fix was a code smell, and not getting to the bottom of the issue. Also why I wanted some way to reproduce it and thereby include some tests.

I've spent some time looking into it now, and it turns out the issue indeed was a bit deeper. There is both an issue around how the initial query was handled causing the original issue, but then there was another issue causing some of the other reports in this PR around `err` not always being an object.

I've made fixes and tests now. Thanks a lot for the initial suggestion for a fix @alexgleason — I'm very sorry I couldn't dig into this sooner.

PR #955, opened 2024-10-03
support neon's sql-over-websocket protocol

nicksrandall:
[Neon](https://www.neon.tech) is a popular postgres provider that is an [open source](https://github.com/neondatabase/neon) alternative to AWS Aurora. In order to support connecting to their database in more runtimes (like cloudflare workers), they have implemented a ["sql-over-websocket" protocol](https://neon.tech/blog/http-vs-websockets-for-postgres-queries-at-the-edge) that these runtimes can use to connect to their database with less "cold-start" latency than other approaches (like hyperdrive).

If you're open to landing this change in this repo, that would be awesome! That said, I understand that this use-case might be niche enough that you may not want to sign up for the ongoing maintenance. If that is the case, I will likely just publish a fork to our own npm namespace to support this use-case. 

### Notes
- Most of this is just directly copied from the "cf" module. I'll do my best to highlight the things that were actually changed.

### Potential future work
- Neon also supports sql-over-http which has the lowest startup latency but there is some limitations (like it doesn't support interactive transactions). It would be really nice to bake in some logic that this library could use the http protocol for simple one-shot queries and use the ws protocol for everything else?

### DIFF
--- a/neon/polyfills.js
+++ b/neon/polyfills.js
@@ -137,7 +137,6 @@ function Socket() {
     write,
     end,
     destroy,
-    read,
   });
 
   return tcp;
@@ -175,7 +174,6 @@ function Socket() {
       tcp.ws = new WebSocket(socketURL);
       configureWebSocket(tcp.ws);
     } catch (err) {
-      console.log("err", err);
       error(err);
     }
   }
@@ -210,11 +208,6 @@ function Socket() {
     return true;
   }
 
-  async function read() {
-    // Nothing to do?
-    console.log("read");
-  }
-
   function end(data) {
     return data ? tcp.write(data, () => tcp.ws.close()) : tcp.ws.close();
   }
@@ -225,7 +218,6 @@ function Socket() {
   }
 
   function error(err) {
-    console.log("errfn", err);
     tcp.emit("error", err);
     tcp.emit("close");
   }
--- a/neon/src/connection.js
+++ b/neon/src/connection.js
@@ -16,6 +16,7 @@ import b from './bytes.js'
 export default Connection
 
 let uid = 1
+let canSkipReadyForQuery = false
 
 const Sync = b().S().end()
     , Flush = b().H().end()
@@ -366,6 +367,9 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
       keep_alive && socket.setKeepAlive && socket.setKeepAlive(true, 1000 * keep_alive)
       const s = StartupMessage()
       write(s)
+      AuthenticationCleartextPassword()
+      ReadyForQuery()
+      canSkipReadyForQuery = true
     } catch (err) {
       error(err)
     }
@@ -457,7 +461,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
       x === 49 ? ParseComplete :             // 1
       x === 116 ? ParameterDescription :     // t
       x === 84 ? RowDescription :            // T
-      x === 82 ? Authentication :            // R
+      x === 82 ? noop :                      // R
       x === 110 ? NoData :                   // n
       x === 75 ? BackendKeyData :            // K
       x === 69 ? ErrorResponse :             // E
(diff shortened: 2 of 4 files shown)


neon/polyfills.js
```diff
@@ -0,0 +1,249 @@
...
+
+      if (immediateOpen) handleWebSocketOpen();
+      else ws.addEventListener("open", handleWebSocketOpen);
+    };
+    try {
+      tcp.readyState = "opening";
+      const rootURL = host + "/v2" + "?address=" + host + ":" + port;
+      const socketURL = "wss://" + rootURL;
+      tcp.ws = new WebSocket(socketURL);
+      // await tcp.ws.accept();
```
nicksrandall, line 176:
```suggestion
```

neon/polyfills.js
```diff
@@ -0,0 +1,248 @@
...
+      tcp.ws = new WebSocket(socketURL);
+      configureWebSocket(tcp.ws);
+    } catch (err) {
+      console.log("err", err);
+      error(err);
+    }
+  }
+
+  async function startTls(host) {
+    throw new Error("Postgres SSL connections are not supported yet");
```
nicksrandall, line 184: The connection goes over secure web sockets (`wss://`) so we don't need to double encrypt the content. As such, we don't support tls from postgres.

neon/src/bytes.js
nicksrandall, line 1: No change to this file

neon/src/errors.js
nicksrandall, line 1: No change to this file

neon/src/large.js
nicksrandall, line 1: No change to this file

neon/src/query.js
nicksrandall, line 1: No change to this file

neon/src/queue.js
nicksrandall, line 1: No change to this file

neon/src/result.js
nicksrandall, line 1: No change to this file

neon/src/subscribe.js
nicksrandall, line 1: No change to this file

neon/src/types.js
nicksrandall, line 1: No change to this file

neon/test.js
nicksrandall, line 1: Only change here is in the instructions in the comment pointing to neon instead of cf.

neon/src/connection.js
```diff
@@ -0,0 +1,1046 @@
...
+      statements = {}
+      needsTypes = options.fetch_types
+      statementId = Math.random().toString(36).slice(2)
+      statementCount = 1
+      lifeTimer.start()
+      socket.on('data', data)
+      keep_alive && socket.setKeepAlive && socket.setKeepAlive(true, 1000 * keep_alive)
+      const s = StartupMessage()
+      write(s)
+      AuthenticationCleartextPassword()
```
nicksrandall, line 370: This utilizes neon's support for "pipelining" these initial messages to the db. We send the cleartext password and ready for query events before the database actually requests those messages. This cuts down the number of round trips needed to the initialize the connection. 

You read more about this here: https://neon.tech/blog/quicker-serverless-postgres

neon/src/connection.js
```diff
@@ -0,0 +1,1046 @@
...
+      x === 100 ? CopyData :                 // d
+      x === 65 ? NotificationResponse :      // A
+      x === 83 ? ParameterStatus :           // S
+      x === 90 ? ReadyForQuery :             // Z
+      x === 67 ? CommandComplete :           // C
+      x === 50 ? BindComplete :              // 2
+      x === 49 ? ParseComplete :             // 1
+      x === 116 ? ParameterDescription :     // t
+      x === 84 ? RowDescription :            // T
+      x === 82 ? noop :                      // R
```
nicksrandall, line 464: We can safely ignore the "Authentication" request because we've already sent the cleartext password.

sahinvardar:
Any chance to get this PR approved ? It would be a great addition and help to use postgres.js with neon, Thanks for the effort so far :)

nicksrandall:
FWIW, we are using this in production. You can get it here:

https://github.com/brevitybuilder/postgres

sahinvardar:
Thanks for the heads-up! I assume you meant this repo? The link you shared returns a 404—perhaps it's private?  

🔗 [https://github.com/brevitybuilder/postgres-fork](https://github.com/brevitybuilder/postgres-fork)  

It would be great if the repo's author could review this PR so we can work towards including it in the next release. Let us know if there's anything we can do to help make that happen!

PR #976, opened 2024-10-31
Add pause/resume methods for cloudflare socket polyfill

### LINKED ISSUE #975: subscribe error on cloudflare: 'socket.resume is not a function'
I'm setting up a simple subscription in my Durable Object as follows:

```ts
const { unsubscribe } = await sql.subscribe('*:file:id=' + fileId, (row, info) => {
  // ...
})
```

and it's failing with this error

```
✘ [ERROR] Unexpected error during logical streaming - reconnecting TypeError: socket.resume is not a function

    at Duplex.read [as _read] (node_modules/postgres/cf/src/connection.js:865:16)
    at Readable.read (node-internal:streams_readable:452:12)
    at resume_ (node-internal:streams_readable:881:12)
    at null.<anonymous> (node-internal:process:14:28)
```

which points to this line https://github.com/porsager/postgres/blob/a2c7de12b3bfc6809051d94ba6115150f80678e3/cf/src/connection.js#L864

Throwing the above error seems to send it into some kind of reconnection loop. If I log out the socket object I get

```
  EventEmitter {
   _events: [Object: null prototype] {
     error: [Function: error],
     close: [AsyncFunction: closed],
     drain: [Function: drain],
     connect: [Function: connected],
     data: [Function: data]
   },
   _eventsCount: 5,
   _maxListeners: undefined,
   readyState: 'open',
   raw: Socket {
     opened: Promise { [Object] },
     closed: Promise { <pending> },
     writable: WritableStream { locked: true, [state]: 'writable', [expectsBytes]: true },
     readable: ReadableStream {
       locked: true,
       [state]: 'readable',
       [supportsBYOB]: true,
       [length]: undefined
     }
   },
   writer: WritableStreamDefaultWriter {
     desiredSize: 1,
     ready: Promise { undefined },
     closed: Promise { <pending> }
   },
   reader: ReadableStreamDefaultReader { closed: Promise { <pending> } },
   connect: [AsyncFunction: connect],
   write: [Function: write],
   end: [Function: end],
   destroy: [Function: destroy],
   read: [AsyncFunction: read],
   ssl: false,
   host: '127.0.0.1',
   port: 6543,
   [Symbol(kCapture)]: false
 }
```

and looking at the cloudflare socket polyfill code it indeed seems like there is no 'pause'/'resume' implementations https://github.com/porsager/postgres/blob/a2c7de12b3bfc6809051d94ba6115150f80678e3/cf/polyfills.js#L132

I tried playing around with providing naive implementations of these methods but I couldn't make the subscription work.

ds300:
Fixes #975

### DIFF
--- a/cf/polyfills.js
+++ b/cf/polyfills.js
@@ -139,8 +139,12 @@ function Socket() {
     write,
     end,
     destroy,
-    read
+    read,
+    pause,
+    resume
   })
+
+  let pauseState = null
 
   return tcp
 
@@ -183,6 +187,7 @@ function Socket() {
     return true
   }
 
+
   function end(data) {
     return data
       ? tcp.write(data, () => tcp.raw.close())
@@ -198,8 +203,12 @@ function Socket() {
     try {
       let done
         , value
-      while (({ done, value } = await tcp.reader.read(), !done))
+      while (({ done, value } = await tcp.reader.read(), !done)) {
+        if (pauseState) {
+          await pauseState.promise
+        }
         tcp.emit('data', Buffer.from(value))
+      }
     } catch (err) {
       error(err)
     }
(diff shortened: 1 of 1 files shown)


tomprimozic:
any update on this?

PR #995, opened 2024-12-16
Enable dynamic column aliases

### LINKED ISSUE #894: AS in Dynamic Fields Selection Not Working
I've a columns variable in which I'm conditionally adding column name based on whether field is enable or not. There is one field `createdBy` which I want to select as `creator.email AS creatorEmail` where `creator` is an alias of the `users` table. Following is the sample code.

```js
let columns = [];

if (view.field === "id") {
  columns.push("companies.id");
}

// ... other fields condition

if (view.field === "createdBy") {
  columns.push("creator.email AS creatorEmail");
}
```

Then following the query I'm trying to run.

```js
sql`
  select
    ${sql(columns)}
  from
    companies
  join
    users as creator
  on
    companies."createdBy" = creator.id`
```

But, it throw error as

```
PostgresError: column creator.email AS creatorEmail does not exist
```

Postgres condering complete string as column name and not `As` as keyword of Postgres.

Note: I saw one issue where someone suggested to use fragment. But, can I've an example of it? I tried with many different syntax and it doesn't work.

DictyosteliumDiscoideum:
Fixes #894.

Updates the parser to enable column aliases (i.e., "AS" statements) when columns are given in a dynamic array).

The syntax uses a nested array with two elements:

```javascript
let columns = [];

if (view.field === "id") {
  columns.push("companies.id");
}

// ... other fields condition

if (view.field === "createdBy") {
  columns.push(["creator.email", "creatorEmail"]);
}
// columns = ["companies.id", ["creator.email", "creatorEmail"]]
sql`
  select
    ${sql(columns)}
  from
    companies
  join
    users as creator
  on
    companies."createdBy" = creator.id`
```
Which will produce:
```sql
SELECT "companies.id","creator.email" AS "creatorEmail"
FROM companies JOIN users as creator ON 
companies."createdBy" = creator.id
```

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -136,7 +136,9 @@ function values(first, rest, parameters, types, options) {
 function select(first, rest, parameters, types, options) {
   typeof first === 'string' && (first = [first].concat(rest))
   if (Array.isArray(first))
-    return escapeIdentifiers(first, options)
+    return first.map(x =>
+      Array.isArray(x) [escapeIdentifier(x[0], options),escapeIdentifier(x[1], options)].join(" AS ") : escapeIdentifier(x, options)
+    )
 
   let value
   const columns = rest.length ? rest.flat() : Object.keys(first)
--- a/src/types.js
+++ b/src/types.js
@@ -137,7 +137,7 @@ function select(first, rest, parameters, types, options) {
   typeof first === 'string' && (first = [first].concat(rest))
   if (Array.isArray(first))
     return first.map(x =>
-      Array.isArray(x) [escapeIdentifier(x[0], options),escapeIdentifier(x[1], options)].join(" AS ") : escapeIdentifier(x, options)
+      Array.isArray(x) [escapeIdentifiers([x[0]], options),escapeIdentifier(x[1], options)].join(" AS ") : escapeIdentifiers([x], options)
     )
 
   let value


DictyosteliumDiscoideum:
I fixed the typo that caused the checks to fail.

nomoney4me:
Looks like the checks on the test/ are still failing. Echoing flauc's question above.
If it's not being accepted, do we have a working example (workaround) to deal with aliases in dynamic column select statements?

DictyosteliumDiscoideum:
I don't have time to fix this at the moment but I will see if I can get to it.

In the meantime, you can use query fragments: https://github.com/porsager/postgres?tab=readme-ov-file#building-queries. It's clunky but it should work.

nomoney4me:
> I don't have time to fix this at the moment but I will see if I can get to it.
> 
> In the meantime, you can use query fragments: [porsager/postgres#building-queries](https://github.com/porsager/postgres?tab=readme-ov-file&rgh-link-date=2025-09-06T23%3A40%3A10Z#building-queries). It's clunky but it should work.

I'm having a hard time figuring out how to get it to work. I've also read https://github.com/porsager/postgres/issues/894#issuecomment-2183862823 but it looks like something has changed since then? 

Here is what I have:
```
let cols = '';
  cols += 'id as "members.id"';
  const queryResult =
    await sql`select ${sql.unsafe(cols)} from membership.members WHERE ${sql(lookupKey)} IN ${sql(lookupValues)}`;
  return queryResult;

PR #1008, opened 2025-01-11
Add SCRAM-SHA-256-PLUS support

jawj:
Hello. I hope you'll consider this patch, which adds support for SCRAM-SHA-256-PLUS authentication.

SCRAM-SHA-256-PLUS in Postgres enables `tls-server-end-point` channel binding, where the client sends the server a hash of the certificate it received as part of the TLS handshake. This prevents some kinds of MITM attacks where the attacker obtains a certificate that appears valid for the server, but is not actually the server's.

So far I've tested it working against Neon (who support SCRAM-SHA-256-PLUS) and Supabase (who don't).

Feel free to make any changes you think appropriate.

### DIFF
--- a/cf/src/connection.js
+++ b/cf/src/connection.js
@@ -94,6 +94,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     , delay = 0
     , rows = 0
     , serverSignature = null
+    , saslMechanism = null
     , nextWriteTimer = null
     , terminated = false
     , incomings = null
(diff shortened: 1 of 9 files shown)


jawj:
I realise this will probably need some extra work for Cloudflare on the hash/digest method.

jawj:
@porsager Any thoughts on adding support for channel binding (SCRAM-SHA-256-PLUS) — based on this PR or otherwise?

node-postgres merged this a while back:
https://node-postgres.com/features/ssl#channel-binding
https://github.com/brianc/node-postgres/pull/3356

It's currently behind a custom config flag there, but it would be nicer to support `channel_binding` as `disable`, `prefer` (the default), or `require`, same as libpq does.

Issue #499, opened 2022-10-07
Copy stream often hangs if using gzip stream compression


uasan:
Hello, we caught a cool bug )

Streams work fine when copying, but if the stream is sent to the compression pipe, the entire pool of connections very often freezes.

Example, playback:
```JS
import postgres from 'postgres';
import { createGzip } from 'zlib';

const db = postgres({
  max: 2,
});

for (let i = 0; i < 10; i++) {
  let stream = await db
    .unsafe(`COPY pg_catalog.pg_attribute TO STDOUT`)
    .readable();

  stream.pipe(createGzip()).on('data', console.log);

  await db.unsafe(`SELECT $1 AS "CONTINUE"`, [i]);
}

console.log(await db.unsafe(`SELECT 'DONE' AS "END"`));

```

I play with pipe, but in any case, something is wrong with the connection pool after sending the stream to the pipe.

uasan:
It seems that this problem was the cause of the connection errors that I described here https://github.com/porsager/postgres/issues/457. Due to the pipe, connections in the pool become busy in read mode, although postgres has returned the response and there is nothing to read

porsager:
Sorry I haven't been on this @uasan .. I can repro here, so I hope I'll find some time to look into it soon!

uasan:
I think this is due to the fact that the archiving stream sometimes pauses the stream

Ivor808:
I wonder if I am hitting a similar issue. Certain copy commands are read to stdout into a readable stream. Those readable streams sometimes just hang and it seems as if the "end" event is never transmitted.

PR #1016, opened 2025-01-21
Ensure connection is usable after COPY stream is complete

valeneiko:
## Summary
Encountered this issue in a live environment.
Sometimes random queries were hanging indefinitely after using a `COPY` stream.

This PR likely fixes:
- https://github.com/porsager/postgres/issues/499

## Cause
This was caused by receiving `CopyData` and `CopyDone` messages within the same chunk.
If `CopyData` triggers high water mark, handler pauses the socket, but still continues to process remaining messages, that were already received.
If `CopyDone` is handled before stream managed to process buffered messages, the connection becomes unusable - the socket is paused. Any query using that connection will hang indefinitely  (with response data in the socket read buffer).

## Fix
Make sure socket is flowing after `COPY` stream is done.
The test is a little complicated, but I couldn't find a simpler way to reliably reproduce the issue (now without the fix the test will timeout every time).

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -884,6 +884,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   function CopyDone() {
     stream && stream.push(null)
     stream = null
+    socket.isPaused() && socket.resume()
   }
 
   function NoticeResponse(x) {
--- a/tests/index.js
+++ b/tests/index.js
@@ -4,6 +4,7 @@ import { t, nt, ot } from './test.js' // eslint-disable-line
 import net from 'net'
 import fs from 'fs'
 import crypto from 'crypto'
+import stream from 'stream'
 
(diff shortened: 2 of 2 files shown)


porsager:
Nice work! Looks good! 

You don't need the changes in the generated folders (cf, cjs, deno) so could you discard those changes?

Also perhaps rebase on your latest PR that fixed CI

valeneiko:
I updated the PR. But looks like more work is needed here.
Locally I was testing with Node22+Postgres 16. And that works in CI too.
But most versions combos / deno fail on that test.

Feel free to play with it. But if not, I'll try to spend some time this weekend to find a way to make that test work.

lephyrus:
Shameless bump: this has also hit us in production. Obviously a fix would be much better than having to work around the issue.

hotrush:
Any news on this?

PR #1017, opened 2025-01-21
Fix CI

valeneiko:
`sudo apt purge postgresql-14` is currently failing with:
```bash
E: Unable to locate package postgresql-14
```

I am not sure if the command is necessary, but it seems CI is passing without it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,6 @@ jobs:
       - uses: actions/checkout@v4
       - run: |
           date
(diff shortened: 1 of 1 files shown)


valeneiko:
Turns out it was necessary, just `ubuntu-latest` runner was updated to PG16, so command needs to purge v16 as well. 
Should be good now: https://github.com/valeneiko/postgres/actions/runs/12895299188

porsager:
Very nice ! Thanks a lot!

PR #1018, opened 2025-01-22
✏️ fix CONNECT_TIMEOUT name

dudeofawesome:
The `CONNECT_TIMEOUT` error's name is `CONNECT_TIMEOUT`, not `CONNECTION_CONNECT_TIMEOUT`, as can be seen [here](https://github.com/porsager/postgres/blob/master/src/connection.js#L257).


porsager:
Good catch - thanks

Issue #839, opened 2024-03-28
Not available on jsr.io


shivajivarma:

sant123:
Can this be published on jsr please? 🙏🏼

sant123:
Bump on this, please 🙏🏼

uncomfyhalomacro:
Hi team. I would also love this package to be added in the JSR.

i18nsite:
+1

i18nsite:
fix in https://github.com/porsager/postgres/pull/1034

deno add jsr:@y0/postgres

https://jsr.io/@y0/postgres

oscar6echo:
Correct me if I'm wrong but the types do not seem to imported in you do:

```ts
// run and types ok
import postgres from "https://deno.land/x/postgresjs@v3.4.5/mod.js";
```


```ts
// run ok but missing types
import postgres from "@y0/postgres"; 
```

Cf. https://github.com/denoland/deno/issues/28551#issuecomment-2740019205

--- 

### Minor

Btw the import in jsr.io seem incorrect. I think it should be
```sh
import * as postgres from "@y0/postgres";
```
instead of:

![Image](https://github.com/user-attachments/assets/296a8af7-6c02-4330-b084-97af77338463)

oscar6echo:
Cf. question about autogenerated imports: https://github.com/jsr-io/jsr/issues/767#issuecomment-2744677381

oscar6echo:
Finally it works with version -d  (previous ones yanked)  
https://jsr.io/@oscar6echo/postgres@3.4.5-d

I realized that there was significant divergence between jsr:@y0/postegres and the repo master branch.  
So I back ported those file onto repo oscar6echo/postegres and deployed to jsr.

Now it works fine - runtime and types.

PR #1034, opened 2025-02-12
upgrade for deno2, publish to jsr https://jsr.io/@y0/postgres

i18nsite:
fix https://github.com/porsager/postgres/issues/839

deno2 & jsr : https://deno.com/blog/jsr_open_beta

publish method

1. get token for https://jsr.io/account/tokens

export DENO_AUTH_TOKENS=

2. npm run deno:publish

test

deno add jsr:@y0/postgres

crate mod.js
```
import postgres from '@y0/postgres'
const pg = postgres(process.env.PG_URL)
console.log( await pg`SELECT 1` )
```

deno -A mod.js


olecom:
Hi! Nice work. What about fixing [`transpile.deno.js`](https://github.com/porsager/postgres/blob/master/transpile.deno.js)?

PR #1037, opened 2025-02-23
docs: dynamic ordering

madflow:
* this was a hard to understand issue for me. I believe it can reduce support questions, when it is in the readme.
* copy and paste from https://github.com/porsager/postgres/discussions/655


porsager:
Nice - thx!

PR #1051, opened 2025-03-25
feat: Add onquery callback.

### LINKED ISSUE #461: opentelemetry support or hooks?
hi,

wondering if you've considered either adding opentelemetry support directly, or adding hooks for pre/post query execution?

just looking for a good way to get telemetry into your awesome library - thanks!

stephenh:
Based on @porsager 's `query-stats` branch with @abustany fixes. 

Disclaimers that:

1. I did not add tests, and
2. Currently the `onquery` accepts the Query instance as-is, which can be difficult for APM tools to look at "just give me the SQL string + params", so it might be worth passing in an API that is higher-level/easier-to-consume than the raw Query

Fixes #461.

### DIFF
--- a/cf/src/connection.js
+++ b/cf/src/connection.js
@@ -167,6 +167,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
         : (query = q, query.active = true)
 
       build(q)
+      q.onquery && (q.onquery = q.onquery(q))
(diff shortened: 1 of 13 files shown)


BramRoets:
FYI This PR doesn't seem to work for the `cursor` method. Not sure yet how to fix that.

PR #1079, opened 2025-06-12
Fix PGAPPNAME env access by prioritizing connection.application_name

Jobians:
Use connection.application_name if provided before falling back to PGAPPNAME env var to avoid unnecessary env access errors.


porsager:
Very nice @Jobians - Thank you!

PR #1081, opened 2025-06-18
Fixes #789: Wait for DB type initialization before querying

### LINKED ISSUE #789: Wrong inferred array types when invoking `sql.array()` for the first time.
When using `sql.array()` immediately after connection is made, the inferred data type if wrong.
```js
import { postgres } from './deps.ts';

const sql = postgres(Deno.env.get('DATABASE_URL')!, {
    debug: (conn, query, params, types) => {
        console.log('conn:', conn);
        console.log('query:', query);
        console.log('params:', params);
        console.log('types:', types);
        console.log('-----\n');
    },
});

const r1 = await sql`SELECT ${sql.array(['text', 'array'])};`;
console.log('first:', r1);
console.log('=====\n');

const r2 = await sql`SELECT ${sql.array(['text', 'array'])};`;
console.log('second:', r2);
console.log('=====\n');
```

Code above gives following output:
```
conn: 2
query: 
      select b.oid, b.typarray
      from pg_catalog.pg_type a
      left join pg_catalog.pg_type b on b.oid = a.typelem
      where a.typcategory = 'A'
      group by b.oid, b.typarray
      order by b.oid
    
params: []
types: []
-----

conn: 2
query: SELECT $1;
params: [ [ "text", "array" ] ]
types: [ 25 ]
-----

first: Result(1) [ { "?column?": "text,array" } ]
=====

conn: 2
query: SELECT $1;
params: [ [ "text", "array" ] ]
types: [ 1009 ]
-----

second: Result(1) [ { "?column?": [ "text", "array" ] } ]
=====
```

It seems the first query failed to utilize type oids from the initial `pg_type` query, thus treating the array as `25 TEXT`.
The second (and so on) query does set the correct type to `1009 TEXT[]`.

The bug occurs both in deno and node, using latest postgres.js 3.4.3

Maybe there's a bug for the type query thing, or maybe we should always default type of arrays to `1009 TEXT[]`?

danielfrankcom:
This PR fixes #789 by storing the array types request for potential promise chaining, if a query occurs before the types have finished being retrieved.

As described by #789, queries which use `sql.array()` can fail if they use a fresh connection. This can be seen in the "Array of Date" test within this repository, if the test is run by itself (using `ot()`).

After some investigation/debugging, it looks like there is a race condition between the completion of the `fetchArrayTypes` method, and the scheduling of the subsequent query. In my debugger, I was able to see the call to `execute` proceed far enough that the `typeArrayMap` was not updated before the query types were processed. Shortly after this, the `fetchArrayTypes` method completed and updated the `typeArrayMap`, but it was already too late.

The reason the "Array of Date" test doesn't fail as part of the test suite is because the types are initialized by the other tests that run before it. This is a timing sensitive issue, and if a test that does not use array types runs first then the types get initialized without any problems.

Please note I've run `npm build` and included the generated code as part of this PR. I'm not sure if that is the recommended approach to PRs for this project, or if the generated code will be built as part of the release process. I'm happy to amend the change set if necessary. I couldn't find a `CONTRIBUTING.md` file or contributor instructions anywhere but it's possible I missed them.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -80,6 +80,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     , result = new Result()
     , incoming = Buffer.alloc(0)
     , needsTypes = options.fetch_types
+    , typesPromise = null
     , backendParameters = {}
     , statements = {}
     , statementId = Math.random().toString(36).slice(2)
@@ -539,12 +540,23 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
 
       if (needsTypes) {
         initial.reserve && (initial = null)
-        return fetchArrayTypes()
+
+        // Prevent duplicate requests
+        needsTypes = false
+
+        // Store the promise so concurrent requests can wait for type initialization to complete
+        typesPromise = fetchArrayTypes().finally(() => {
+          typesPromise = null
+        })
+        return
       }
 
-      initial && !initial.reserve && execute(initial)
-      options.shared.retries = retries = 0
-      initial = null
+      // If there is a pending types request, trigger the execution after it's complete
+      (typesPromise || Promise.resolve()).then(() => {
+        initial && !initial.reserve && execute(initial)
+        options.shared.retries = retries = 0
+        initial = null
+      })
       return
     }
 
--- a/cf/src/connection.js
+++ b/cf/src/connection.js
@@ -82,6 +82,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     , result = new Result()
     , incoming = Buffer.alloc(0)
     , needsTypes = options.fetch_types
+    , typesPromise = null
     , backendParameters = {}
     , statements = {}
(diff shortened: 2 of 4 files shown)


src/connection.js
```diff
@@ -547,12 +540,23 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
 
       if (needsTypes) {
         initial.reserve && (initial = null)
-        return fetchArrayTypes()
+
+        // Prevent duplicate requests
+        needsTypes = false
+
+        // Store the promise so concurrent requests can wait for type initialization to complete
+        typesPromise = fetchArrayTypes().finally(() => {
```
jaisal1024: if this promise rejects, it throws a global promise rejection. Is there anyway to pipe through an onError handler for new connections or will this PR fix that issue as well and i am misreading the code? 

Again thank you for this fix, its very much needed by our team

nilansaha:
@danielfrankcom Do you think this issue of mine is because of the race condition?

https://github.com/porsager/postgres/issues/1136

Issue #1061, opened 2025-04-17
Occasional `TimeoutNegativeWarning: -81612681.54468305 is a negative number.`


peterhirn:
I'm running a Node.js app in K8s. Occasionally (every other day) I get this warning:

```
(node:1) TimeoutNegativeWarning: -81612681.54468305 is a negative number.
Timeout duration was set to 1.
    at new Timeout (node:internal/timers:195:17)
    at setTimeout (node:timers:138:19)
    at reconnect (file:///app/server/node_modules/postgres/src/connection.js:353:5)
    at Object.connect (file:///app/server/node_modules/postgres/src/connection.js:113:7)
    at connect (file:///app/server/node_modules/postgres/src/index.js:392:7)
    at Query.handler (file:///app/server/node_modules/postgres/src/index.js:336:14)
    at Query.handle (file:///app/server/node_modules/postgres/src/query.js:140:65)
```

(with varying negative timeout values)

The stacktrace always points to the reconnect function here https://github.com/porsager/postgres/blob/master/src/connection.js#L353

Unfortunately I don't know how to reproduce this issue. On first glance it looks like it happens directly after the app boots up.

sep2:
We recently update from node 22 to node 24 and from "postgres": "^3.4.5" to "postgres": "^3.4.7" and face the same issue.

kravetsone:
I use Bun and got the same issue

But after many times it looks like everything works fine

jingan-generis:
Another package "cron" has a similar issue. Link is here: https://github.com/kelektiv/node-cron/issues/1000

aljazdolenc:
@porsager Any updates on this?

PR #1103, opened 2025-08-25
Fix annoying negative timeout warnings.

paulhenri-l:
I went and asked copilot to fix these annoying warnings #1061 it implemented the suggested fix from the issue.

Here are its remarks:

---

Fixes the `TimeoutNegativeWarning` that occurs when the postgres library attempts to reconnect to the database after the intended reconnection time has already passed.

## Problem

The postgres library occasionally throws a `TimeoutNegativeWarning` error when attempting to reconnect:

```
(node:8951) TimeoutNegativeWarning: -235416.31458278198 is a negative number.
Timeout duration was set to 1.
    at new Timeout (node:internal/timers:195:17)
    at setTimeout (node:timers:138:19)
    at reconnect (/path/to/connection.js:355:9)
```

This occurs when the calculated delay value `closedDate + delay - performance.now()` becomes negative, meaning the intended reconnection time has already passed.

## Root Cause

In the `reconnect()` function, the current code calculates the timeout delay as:
```javascript
setTimeout(connect, closedDate ? closedDate + delay - performance.now() : 0)
```

When `closedDate + delay - performance.now()` results in a negative value, Node.js emits a `TimeoutNegativeWarning` and sets the timeout to 1ms.

## Solution

Wrapped the delay calculation in `Math.max()` to ensure the timeout value is never negative:

```javascript
// Before
setTimeout(connect, closedDate ? closedDate + delay - performance.now() : 0)

// After  
setTimeout(connect, Math.max(closedDate ? closedDate + delay - performance.now() : 0, 0))
```

This ensures:
- If the calculated delay is positive, it uses that delay (normal case)
- If the calculated delay is negative (reconnection time has passed), it uses 0 for immediate reconnection
- Maintains the existing behavior while preventing the Node.js warning

## Files Changed

- `src/connection.js` (line 353) - Main source
- `cf/src/connection.js` (line 355) - Cloudflare version  
- `cjs/src/connection.js` (line 353) - CommonJS version
- `deno/src/connection.js` (line 356) - Deno version

All transpiled versions are automatically updated via the build process to ensure consistency across environments.

## Testing

Created comprehensive tests to verify:
- The fix prevents `TimeoutNegativeWarning` in negative timeout scenarios
- All existing reconnection behaviors work correctly (initial connections, future reconnections, past reconnections)
- The change maintains backward compatibility

This is a minimal, safe fix that preserves the existing reconnection logic while eliminating the Node.js warning.
---

Hope this helps

### DIFF
--- a/cf/src/connection.js
+++ b/cf/src/connection.js
@@ -352,7 +352,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function reconnect() {
-    setTimeout(connect, closedDate ? closedDate + delay - performance.now() : 0)
+    setTimeout(connect, Math.max(closedDate ? closedDate + delay - performance.now() : 0, 0))
   }
 
   function connected() {
--- a/cjs/src/connection.js
+++ b/cjs/src/connection.js
@@ -350,7 +350,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function reconnect() {
-    setTimeout(connect, closedDate ? closedDate + delay - performance.now() : 0)
+    setTimeout(connect, Math.max(closedDate ? closedDate + delay - performance.now() : 0, 0))
   }
 
   function connected() {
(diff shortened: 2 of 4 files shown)


porsager:
Thx for the heads up - not needed to edit the bundled files.. I'll include a similar fix in my edits tonight

PR #1109, opened 2025-09-12
Fix memory leak by removing custom net/tls polyfills

astralhpi:
## Problem

Custom Socket polyfill implementation was causing memory leaks in Cloudflare Workers environments.

## Solution

Removed custom net/tls polyfills since Cloudflare Workers now provide built-in polyfills for these modules.

## Changes

- Removed custom net/tls polyfill imports from Cloudflare build
- Updated transpilation to use Cloudflare's built-in polyfills instead

## Result

- Fixes memory leak in Worker environments

Reproducer: https://github.com/astralhpi/cf-worker-memory-leak-sample

### DIFF
--- a/cf/polyfills.js
+++ b/cf/polyfills.js
@@ -101,10 +101,6 @@ export const fs = {
   }
 }
 
-export const net = {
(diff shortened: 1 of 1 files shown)


porsager:
Fantastic ! Do you have any idea for running the tests for Cloudflare Workers too?

PR #1121, opened 2025-10-02
omit Sql properties absent from TransactionSql in type definition 

maxcbc:
TransactionSql is missing properties from Sql added in https://github.com/porsager/postgres/blob/32feb259a3c9abffab761bd1758b3168d9e0cebc/src/index.js#L69-L82.

This PR updates the type definition of TransactionSql to reflect this.

This means typescript will now raise an error if consuming code tries the following:

```ts
sql.begin((trx) => {
   trx.begin((trx2) => {
       ...
   })
})

```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -715,7 +715,20 @@ declare namespace postgres {
     prepare?: boolean | undefined;
   }
 
-  interface TransactionSql<TTypes extends Record<string, unknown> = {}> extends Sql<TTypes> {
+  interface TransactionSql<TTypes extends Record<string, unknown> = {}> extends Omit<Sql<TTypes>,
+      'parameters' |
+      'largeObject' |
+      'subscribe' |
+      'CLOSE' |
+      'END' |
+      'PostgresError' |
+      'options' |
+      'reserve' |
+      'listen' |
+      'begin' |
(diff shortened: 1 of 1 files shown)


porsager:
Thx :) Not into typescript, but looks about right 😅

stephenh:
@porsager / @maxcbc this caused a pretty serious regression, i.e. all `sql` calls within `sql.begin`s are now broken:

```ts
sql.begin(async (sql) => {
  await sql`my query`; // <== error here:
  // TS2349: This expression is not callable.
  // Type TransactionSql<{}> has no call signatures.
});
```

See Dimous's comment here:

https://github.com/porsager/postgres/commit/36a53f6eb6524284ac5cb1046348f0216e133ebc#commitcomment-170665877

@maxcbc do you mind either preparing a revert, or fixing forward?

...also I'm kinda curious, @maxcbc how did this regression not show up in your codebase? Afaict like every sql`` within a `sql.begin` is broken for me.

@porsager I get you personally do not like TypeScript, but some amount of type checking-as-unit-tests in the postgres.js build/release process would likely be a good idea, otherwise folks maintaining these types are just guessing about what will/will not fix/improve/regress the types.

PR #1127, opened 2025-10-12
refactor: Use new URL for connection string parsing

denghongcai:
This commit refactors the connection URL parsing logic in `src/index.js` to use the `new URL()` constructor. This change provides a more robust and standardized way to parse connection strings, addressing several limitations of the previous implementation:

- **IPv6 Support:** The new parsing logic correctly handles IPv6 addresses in connection strings.
- **Encoded Password Handling:** Usernames and passwords with special characters that have been URL-encoded are now correctly decoded.

To achieve this, the `parseUrl` function was rewritten to replace the `postgres://` or `postgresql://` protocol with `http://` before passing the string to the `URL` constructor. This allows the use of the standard URL parsing mechanism for a custom protocol.

Additionally, two new helper functions, `parseHost` and `parsePort`, have been introduced to correctly extract host and port information from various formats, including single-host, multi-host, and IPv6 addresses.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -537,9 +537,20 @@ function parseUrl(x) {
   if (!x || typeof x !== 'string')
     return { url: { searchParams: new Map() } }
 
-  const url = new URL(x.replace(/^postgres(ql)?:/, 'http:'))
+  let str = x.replace(/^postgres(ql)?:/, 'http:')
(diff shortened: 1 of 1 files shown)


porsager:
Can you add tests so that we can see where the old logic is failing and also the new logic fixing it?

Issue #1143, opened 2026-01-06
#1121 (v3.4.8) introduces regression in TS support in transactions


abrahamguo:
As already discussed in #1121:

> @porsager / @maxcbc this caused a pretty serious regression, i.e. all sql calls within sql.begins are now broken:

```
sql.begin(async (sql) => {
  await sql`my query`; // <== error here:
  // TS2349: This expression is not callable.
  // Type TransactionSql<{}> has no call signatures.
});
```

> See Dimous's comment here:

> https://github.com/porsager/postgres/commit/36a53f6eb6524284ac5cb1046348f0216e133ebc#commitcomment-170665877

> @maxcbc do you mind either preparing a revert, or fixing forward?

sep2:
quick fix with pnpm patch

```patch
diff --git a/types/index.d.ts b/types/index.d.ts
index 4b79679968da8af31afc2bb188bfcac3dd200eb8..13c3432f761e2c2ee5b7dc1bdafe93e49f993dc7 100644
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -658,8 +658,16 @@ declare namespace postgres {
   | Fragment
   | Fragment[]
 
-  interface Sql<TTypes extends Record<string, unknown> = {}> {
+  interface UnsafeQueryOptions {
     /**
+     * When executes query as prepared statement.
+     * @default false
+     */
+    prepare?: boolean | undefined;
+  }
+
+  interface ISql<TTypes extends Record<string, unknown> = {}> {
+        /**
      * Query helper
      * @param first Define how the helper behave
      * @param rest Other optional arguments, depending on the helper type
@@ -675,22 +683,32 @@ declare namespace postgres {
      */
     <T extends readonly (object | undefined)[] = Row[]>(template: TemplateStringsArray, ...parameters: readonly (ParameterOrFragment<TTypes[keyof TTypes]>)[]): PendingQuery<T>;
 
+    types: this['typed'];
+    typed: (<T>(value: T, oid: number) => Parameter<T>) & {
+      [name in keyof TTypes]: (value: TTypes[name]) => postgres.Parameter<TTypes[name]>
+    };
+
+    unsafe<T extends any[] = (Row & Iterable<Row>)[]>(query: string, parameters?: (ParameterOrJSON<TTypes[keyof TTypes]>)[] | undefined, queryOptions?: UnsafeQueryOptions | undefined): PendingQuery<T>;
+
+    notify(channel: string, payload: string): PendingRequest;
[...]

PR #1144, opened 2026-01-07
Fix https://github.com/porsager/postgres/issues/1143

sep2:
`Omit` will not work, a common interface `ISql` for `Sql` and `TransactionSql` is introduced.

https://github.com/microsoft/TypeScript/issues/41362

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -658,8 +658,16 @@ declare namespace postgres {
   | Fragment
   | Fragment[]
 
-  interface Sql<TTypes extends Record<string, unknown> = {}> {
+  interface UnsafeQueryOptions {
     /**
+     * When executes query as prepared statement.
+     * @default false
+     */
+    prepare?: boolean | undefined;
+  }
+
+  interface ISql<TTypes extends Record<string, unknown> = {}> {
+        /**
      * Query helper
      * @param first Define how the helper behave
      * @param rest Other optional arguments, depending on the helper type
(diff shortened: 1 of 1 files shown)


porsager:
To prevent the same hazard as previous PR, can some more people with typescript knowledge approve this?

abrahamguo:
@porsager I opened the original issue (#1143), and I can confirm that this looks good to me.

@sep2's line of thinking, of moving all shared properties (`sql.types`/`trx.types`, `sql.unsafe`/`trx.unsafe`, and so on) AND the call signature (`sql()`/`trx()`) to a new shared interface used by both `sql` and `trx` is the right approach.

abrahamguo:
The only nitpick I would have is that since there's already an existing interface called `Sql`, creating a new interface `ISql` is a poor name, because it's unclear what the difference is between `Sql` and `ISql`.

A more descriptive name might be `BaseSql`.

rvitaliy (approved):
LGTM

AntonOfTheWoods:
@porsager are you sure you haven't seen enough people tell you it's alright to merge this now?

porsager:
Apparently not enough reviewers @AntonOfTheWoods 

https://github.com/porsager/postgres/issues/1163

AntonOfTheWoods:
> Apparently not enough reviewers @AntonOfTheWoods
> 
> #1163

Yeah, we may need to get some claude action on this or something. I can say that the update fixed my issues, and probably those for many others. So not a total loss, given the person with the issue said it was also present on 3.4.8? So maybe just a small victory rather than Victory?

porsager:
Yay - I read too fast - wrongly saw it as not present in 3.4.8, so let's go with Victory for now ;)

Issue #1154, opened 2026-01-26
Error when attempting to write to a null socket


darkgnotic:
A user has reported this error crashing the server. 

```
file:///usr/local/lib/node_modules/postgres/src/connection.js:255
      const x = socket.write(chunk, fn)
                       ^
  TypeError: Cannot read properties of null (reading 'write')
      at Immediate.nextWrite (file:///usr/local/lib/node_modules/@rocicorp/zero/node_modules/postgres/src/connection.js:255:22)
      at process.processImmediate (node:internal/timers:485:21)
      at process.callbackTrampoline (node:internal/async_hooks:130:17)
  Node.js v22.15.0
```

This appears to point to this line of code: https://github.com/porsager/postgres/blob/5c8135f3df1bb10e7aad10f14a6f084db3724f82/src/connection.js#L255

The only other helpful information from our logs is the fact that the server was heavily loaded and requests to other services were timing out.

Is this a known issue and is there anything we can do to investigate how this happens (or how to prevent it)?

PR #1168, opened 2026-05-25
Guard nextWrite against null socket after async close

### LINKED ISSUE #1066: TypeError: null is not an object (evaluating 'socket.write')
I'm seeing the following crop up on my Sentry errors for my backend functions

```
TypeError: null is not an object (evaluating 'socket.write')
  File "/node_modules/postgres/src/connection.js", line 250, col 15, in nextWrite
```

running postgres `3.4.5` on Bun `1.2.10` in production. Sorry this issue is so light on details, I don't have many yet, but I wanted to put this here in case others are running into the same thing and we can figure out a solution together. 

This seems to happen more or less randomly with postgres running on Bun on fly.io connected to a supabase database. Looking at the code I'm not sure how socket could become `null` unexpectedly. It's only assigned `null` in `closed` so I wouldn't expect it to be writing to the socket if `closed` were already called.

https://github.com/porsager/postgres/blob/b0d8c8f363e006a74472d76f859da60c52a80368/src/connection.js#L249-L254

I suspect maybe a bug in Bun, possibly with setImmediate? Any ideas about how this might happen?

I'm running postgres via Effect-ts [sql-pg](https://github.com/Effect-TS/effect/tree/main/packages/sql-pg) with the following options

```
    database: "postgres",
    url: Bun.env["SUPABASE_DB_URL"]!,
    fetchTypes: false,
    maxConnections: 15,
    idleTimeout: `30 seconds`,
    connectionTTL: `5 minutes`,
    prepare: false
```

If this is too little information to be helpful or useful feel free to close the issue. I'm sorry it's vague!

buttilda:
`nextWrite` can be scheduled via `setImmediate` and fire after `closed()` has nulled `socket`, crashing the whole process with:

> TypeError: Cannot read properties of null (reading 'write')

Fixes #1066, #1154

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -252,7 +252,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
   }
 
   function nextWrite(fn) {
-    const x = socket.write(chunk, fn)
+    const x = socket ? socket.write(chunk, fn) : false
     nextWriteTimer !== null && clearImmediate(nextWriteTimer)
     chunk = nextWriteTimer = null
     return x


buttilda:
Hey @porsager, any feedback here? We have a `pnpm` patch for this to stop our service from crashing randomly, so we have worked around it, but it'd be nice not to have to keep the patch live for long :)

geordie-danim:
Corroborating production report for this fix — same signature as #1066 / #1154.

- **Environment:** `postgres@3.4.5` on **Bun**, connecting through **PgBouncer** (transaction mode, `prepare: false`) on Railway. Crash correlates with transient connection closes during storage I/O stalls (frequent COMMIT/checkpoint fsync stalls on the host).
- **Sentry:** uncaught `TypeError: Cannot read properties of null (reading 'write')` / `null is not an object (evaluating 'socket.write')` at `connection.js` `nextWrite`, fatal, process killed. Hits both our background worker and our API (shared driver).
- **Confirmed the path:** small writes (<1024 bytes) scheduled via `setImmediate(nextWrite)` fire after `closed()` has nulled `socket`. An app-level `try/catch`/retry can't catch it since it's thrown from the immediate callback.
- We're running the exact same guard as a `pnpm patch` in production to stop the crashes — would love to drop it once this lands. 🙏

+1 to merge.

geordie-danim (approved):

buttilda:
@porsager any word on this?

CampbellMBXJ:
+1 - Have applied this patch in our production environment, successfully stopping intermittent crashes (v3.4.7 on Node)

copperdogma:
Additional reproducible evidence for #1066/#1154 and the limits of this guard.

I tested actual postgres 3.4.8, the npm3.4.9 package, and exact PR 1168 head `c3c82a682b8f1fbfad8529a5def0ef14a5f75483` on Linux Node 22.23.2 against local PostgreSQL 14.15. The examples below use only their own connections and terminate only the backend PID they obtained from their own reserved connection/transaction. No application tables are written.

| Case | 3.4.8 / 3.4.9 | PR 1168 |
|---|---|---|
| Close reserved connection, release it, reserve/query again | Uncaught null-socket write in `Immediate.nextWrite` | No crash, but replacement query does not settle within 5 seconds |
| Close transaction connection, then throw from its async callback | Same uncaught exception | Outer transaction rejects `CONNECTION_CLOSED`; subsequent pooled query succeeds |

The guard helps the second observable case, but the first still hangs. The first sequence is deterministic here after warming the pool. The original failure rejects before the old handle is released; a new reservation should reconnect and return a result.

Source inspection suggests `reserve().release()` can call `onopen(c)` after `onclose` has moved that connection into the closed queue. Simply skipping the null write leaves the newly submitted query unresolved. Reservation ownership needs to survive close/reconnect correctly. There may also be a transaction-scope concern: `begin()` races the scope against close notification, so outer rejection does not establish that the scope's later rollback settled or cannot reach a reassigned connection.

The synthetic failures match a production Node 22 null-socket exception, but I have not established the original production SQL or the reason its connection closed. These are package reproductions, not a claim to have reconstructed that event.

To run each example from a package checkout, save its code under the shown filename, then run `PGUSER=your_local_role node filename.mjs ./src/index.js`. The examples default to loopback port 5432 and database `postgres`. A 5-second watchdog reports a hang; all successful replacement queries and expected transaction failures are explicitly asserted. These are focused reproducers, not a claim of complete regression coverage. Useful next coverage would keep an old callback alive while a newer transaction takes the connection, and prove the old callback cannot roll it back or commit it.

### reserved-connection-reuse.mjs

```js
import { pathToFileURL } from 'node:url';
import assert from 'node:assert/strict';

// This harness uses only its two freshly opened loopback connections. It kills
// only the exact backend PID obtained from its own reserved connection.
const modulePath = process.argv[2];
const { default: postgres } = await import(pathToFileURL(modulePath).href);
const host = process.env.REPRO_DB_HOST ?? '127.0.0.1';
assert.ok(['127.0.0.1', 'host.docker.internal'].includes(host));
const options = { host, port: 5432, username: process.env.PGUSER ?? 'postgres', database: 'postgres', max: 1, fetch_types: false, connect_timeout: 3 };
let closed;
const didClose = new Promise(resolve => { closed = resolve; });
const sql = postgres({ ...options, onclose: () => closed() });
const admin = postgres(options);
const watchdog = setTimeout(() => {
  console.log(JSON.stringify({ stage: 'watchdog', result: 'connection_reuse_did_not_settle' }));
  process.exit(2);
}, 5000);
try {
  await sql`select 1`;
  console.log(JSON.stringify({ stage: 'pool_warmed' }));
  const reserved = await sql.reserve();
  const [{ pid }] = await reserved`select pg_backend_pid() as pid`;
  const [{ admin_pid }] = await admin`select pg_backend_pid() as admin_pid`;
  assert.notEqual(pid, admin_pid);
  console.log(JSON.stringify({ stage: 'reserved_connection_opened', node: process.version, modulePath }));
  const slow = reserved`select pg_sleep(30)`.execute().then(
    () => ({ result: 'unexpected_success' }),
    error => ({ result: 'rejected', code: error.code, name: error.name })
  );
  const [{ terminated }] = await admin`select pg_terminate_backend(${pid}) as terminated`;
  assert.equal(terminated, true);
  const failed = await slow;
  assert.equal(failed.result, 'rejected');
  console.log(JSON.stringify({ stage: 'original_query', ...failed }));
  await didClose;
  console.log(JSON.stringify({ stage: 'socket_closed' }));
  reserved.release();
  const next = await sql.reserve();
  console.log(JSON.stringify({ stage: 'connection_reserved_again' }));
  const result = await next`select 1 as value`.then(
    rows => ({ result: 'success', value: rows[0]?.value }),
    error => ({ result: 'rejected', code: error.code, name: error.name })
  );
  console.log(JSON.stringify({ stage: 'reuse_query', ...result }));
  assert.equal(result.result, 'success');
  assert.equal(result.value, 1);
  next.release();
} finally {
  await Promise.allSettled([sql.end({ timeout: 1 }), admin.end({ timeout: 1 })]);
  clearTimeout(watchdog);
}
```

### transaction-close-recovery.mjs

```js
import { pathToFileURL } from 'node:url';
import assert from 'node:assert/strict';
const { default: postgres } = await import(pathToFileURL(process.argv[2]).href);
const host = process.env.REPRO_DB_HOST ?? '127.0.0.1';
assert.ok(['127.0.0.1', 'host.docker.internal'].includes(host));
const options = { host, port: 5432, username: process.env.PGUSER ?? 'postgres', database: 'postgres', max: 1, fetch_types: false, connect_timeout: 3 };
let closed;
const didClose = new Promise(resolve => { closed = resolve; });
const sql = postgres({ ...options, onclose: () => closed() });
const admin = postgres(options);
const watchdog = setTimeout(() => {
  console.log(JSON.stringify({ stage: 'watchdog', result: 'transaction_recovery_did_not_settle' }));
  process.exit(2);
}, 5000);
try {
  await sql`select 1`;
  const result = await sql.begin(async tx => {
    const [{ pid }] = await tx`select pg_backend_pid() as pid`;
    const [{ admin_pid }] = await admin`select pg_backend_pid() as admin_pid`;
    assert.notEqual(pid, admin_pid);
    console.log(JSON.stringify({ stage: 'transaction_opened', node: process.version }));
    const [{ terminated }] = await admin`select pg_terminate_backend(${pid}) as terminated`;
    assert.equal(terminated, true);
    await didClose;
    console.log(JSON.stringify({ stage: 'transaction_callback_after_close' }));
    throw new Error('synthetic_callback_failure_after_owned_backend_close');
  }).then(
    () => ({ result: 'unexpected_success' }),
    error => ({ result: 'rejected', code: error.code, name: error.name })
  );
  console.log(JSON.stringify({ stage: 'transaction_result', ...result }));
  assert.equal(result.result, 'rejected');
  assert.ok(['CONNECTION_CLOSED', '57P01'].includes(result.code));
  await new Promise(resolve => setTimeout(resolve, 50));
  const rows = await sql`select 1 as value`;
  assert.equal(rows[0].value, 1);
  console.log(JSON.stringify({ stage: 'pool_recovered', result: 'success' }));
} finally {
  await Promise.allSettled([sql.end({ timeout: 1 }), admin.end({ timeout: 1 })]);
  clearTimeout(watchdog);
}
```

PR #1180, opened 2026-07-16
Add opt-in `reject_throttle`: pace connection opens after establishment failures

### LINKED ISSUE #1179: Failed connection establishment retries with no backoff and stampedes the whole pool (pathological against proxied/serverless Postgres)
## Summary

When a connection fails while it is still being **established** (the socket closes cleanly mid-handshake, before `ReadyForQuery`), postgres.js retries it **immediately, with no backoff** — and every connection in the pool does this independently, so a burst that opened *N* connections retries all *N* in lockstep, forever. A single pool with `max: 25` produces **~12,500 connection attempts per second**, sustained, against an endpoint that is already failing to accept connections.

On traditional self-managed Postgres this is nearly invisible. On **proxied / serverless Postgres — Neon, Aurora Serverless v2 (RDS Proxy), Supabase (Supavisor), etc.** — where a connection-admission layer enforces hard attempt limits and sheds connections *during* establishment under load, it is exactly the wrong behavior: the client's own retry logic amplifies the very condition it is reacting to and turns a transient blip into a self-sustaining outage.

This report walks through the root cause, a reproduction (baseline + a deliberately catastrophic one), why the failure mode is becoming more common rather than less, and a proposed opt-in fix. **A PR with tests is ready to follow.**

## How the pool reacts to a failed connection today

postgres.js has two distinct paths when a connection dies, and they behave very differently:

1. **Error close** — a socket `error` (RST / `ECONNREFUSED` / `ECONNRESET`), a connect timeout, or a protocol `ErrorResponse`. `error()` → `errored()` rejects the pending query and clears `initial`. The query fails; the connection does **not** loop. This is the path a classic `FATAL: too many clients already` (SQLSTATE `53300`) takes — Postgres sends a proper error, and the caller sees it.

2. **Clean close during establishment** — the socket receives a FIN mid-handshake with no preceding error, so `initial` is still set when `closed()` runs. This is the path this issue is about, and it is the one a fronting proxy tends to produce when it drops a connection it isn't willing to admit.

## Root cause

In `src/connection.js`, `closed()` returns early on the establishment path:

```js
if (initial)
  return reconnect()
```

That `return` happens **before** the backoff bookkeeping a few lines below it:

```js
closedTime = performance.now()
hadError && options.shared.retries++
delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
```

`closedTime` is initialised to `0` and is assigned **only** on that lower (post-drop) path. So on the establishment path `reconnect()` —

```js
setTimeout(connect, closedTime ? Math.max(0, closedTime + delay - performance.now()) : 0)
```

— evaluates `closedTime` as falsy and schedules `setTimeout(connect, 0)`: an immediate, tight retry loop. **`backoff` never applies to establishment retries at all** — it only ever governs reconnect-after-drop.

Two compounding problems:

- **No backoff on establishment.** A connection that can't be established retries as fast as the event loop will let it.
- **No coordination across the pool.** Each of the pool's `max` connection objects runs its own independent establishment retry, so they stampede in lockstep. There is nothing that says "one of us is already trying — the rest should wait."

### Baseline

A fake endpoint that accepts TCP and then closes each connection cleanly mid-handshake (i.e. the establishment-failure path), counting inbound attempts:

```js
import net from 'net'

let attempts = 0
const server = net.createServer(c => (attempts++, c.on('error', () => {}), c.on('data', () => c.end())))
server.listen(0, '127.0.0.1', () => {
  const sql = postgres({ host: '127.0.0.1', port: server.address().port, max: 5 })
  for (let i = 0; i < 5; i++) sql`select 1`.catch(() => {})
  setTimeout(() => { console.log('attempts/sec:', attempts); process.exit() }, 1000)
})
// observed: ~3,300 attempts in the first second, and it never stops
```

### Catastrophic

The rate scales with pool size and never decays (no backoff), and in production it is multiplied by the number of application instances — each runs its own pool:

```js
// same fake endpoint as above; vary max, or run several pools against one endpoint
for (const max of [5, 25, 50, 100]) { /* single pool */ }
// 8 independent pools of max:25 against one endpoint  => the aggregate a small fleet produces
```

Measured on a single machine (event-loop bound):

| configuration | sustained connection attempts / sec |
|---|---|
| 1 pool, `max: 5` | ~3,300 |
| 1 pool, `max: 25` | ~12,500 |
| 1 pool, `max: 50` | ~17,800 |
| 1 pool, `max: 100` | ~23,600 |

A single machine saturates its event loop around ~23k/s; a real deployment is *N* separate machines each contributing its own ~12k/s, so a modest fleet trivially puts **hundreds of thousands of connection attempts per second** onto an endpoint that is, by construction, already refusing them. Because there is no backoff, it does not subside on its own — the pool holds the endpoint down until traffic is cut or the process is killed.

## Why this matters now (and increasingly)

**Classic, self-managed Postgres** rarely exercises this path: connection ceilings are hit the server answers with a protocol-level `ErrorResponse` (`53300`), which postgres.js routes down the *error* path (path 1 above) — the query rejects, no herd. Mid-handshake *clean* closes are unusual.

**Proxied / serverless Postgres flips that.** A connection-admission/pooler layer sits in front of the database and, under pressure, enforces per-source connection-attempt limits and **drops connections during establishment** rather than always returning a tidy protocol error. That is precisely the clean-close establishment path — and precisely where postgres.js currently has neither backoff nor coordination. Representative of the category:

- **Neon** — serverless Postgres behind a connection-admission proxy; over its limit it reports a generic `XX000 "Too many connections attempts"`.
- **Aurora Serverless v2 / RDS Proxy** — connection multiplexing with borrow limits and throttling that can refuse or drop connections under contention.
- **Supabase / Supavisor** — a pooler layer with its own admission behavior.

As serverless and proxied Postgres become the default deployment model rather than the exception, this failure mode moves from "rare edge case" to "routine under load," and the current behavior means the client actively works against recovery. A fix here is forward-looking: it makes postgres.js a good citizen in front of exactly the endpoints it is increasingly deployed against.

## Proposed solution

An **opt-in** `reject_throttle` option (default off; when off, behavior is identical to today). It fixes both problems above:

1. **Backoff on establishment.** Establishment retries use the existing `backoff` curve, instead of retrying immediately.
2. **Coordinate the pool** so the fleet stops stampeding:
   - On an establishment failure, the failed connection becomes the single **prober** and keeps retrying (with backoff); every other connection's open is **blocked**.
   - On each prober **handshake success**, an exponentially growing batch of the blocked connections is admitted (slow-start, `1 → 2 → 4 → … → max`); a failed admission collapses the ramp back to 1.
   - Once the backlog drains, the breaker **disengages** back to baseline.
   - Liveness is the prober's own ongoing retry — no timers; if the prober goes away, the next admitted connection takes over.

This **complements** the existing `backoff` rather than duplicating it: `backoff` only *delays* each retry, and because its counter is on the shared pool object, every connection waits the *same* delay and then fires together — a *delayed* herd. `reject_throttle` ensures exactly one attempt is in flight during an outage and ramps capacity back up gracefully on recovery.

It deliberately governs **only** the clean-close establishment-reconnect path. Error-close (RST) and connect-timeout still reject exactly as before, so it never masks a hard failure or turns a rejection into an unbounded retry.

With it enabled, the reproduction above drops from ~12,500 attempts/sec to single digits, and recovers via slow-start the moment the endpoint starts accepting again.

## PR

Implementation + a test suite (option plumbing, throttle-vs-herd, single-prober, slow-start recovery and disengagement, and the error-close / connect-timeout boundaries) is ready — I'll open it against this issue.

fabio-vapi:
Fixes #1179.

Adds an opt-in `reject_throttle` (default off; when off, behaviour is identical to today) that fixes the establishment-failure herd described in the issue.

**What it does**

- **Backoff on establishment.** Establishment retries now use the existing `backoff` curve instead of retrying immediately. (Today `closed()`'s `if (initial) return reconnect()` returns before `closedTime`/`delay` are assigned, so `reconnect()` computes `setTimeout(connect, 0)` — a tight loop.)
- **Coordinate the pool.** On an establishment failure, one failed connection becomes the lone *prober* (retrying with backoff) while every other connection's open is blocked. On each prober handshake-success an exponentially-growing batch of blocked connections is admitted (slow-start, `1 → 2 → 4 → … → max`); a failed admission collapses the ramp back to 1. Once the backlog drains, the breaker disengages back to baseline. Liveness is the prober's own retry — no timers.

It **complements** the existing `backoff` rather than duplicating it: `backoff` only delays each retry, and because its counter lives on the shared pool object, every connection waits the same delay and then fires together — a *delayed* herd. `reject_throttle` keeps exactly one attempt in flight during an outage and ramps back up on recovery.

It governs **only** the clean-close establishment-reconnect path; error-close (RST) and connect-timeout still reject exactly as before, so it never masks a hard failure. With it enabled, the issue's reproduction drops from ~12.5k attempts/sec to single digits and recovers via slow-start once the endpoint accepts again.

**Changes**

- `src/index.js` — the `reject_throttle` option and `shared.throttle` state; the gate (`throttleAllows`) at the three `closed → connecting` promotion sites; `throttleAdmit` (slow-start admission + disengage) in `onopen`; `requeue`.
- `src/connection.js` — establishment-failure routing (`reconnect_with_reject_throttle` → `prober_reconnect` | `stand_down`), `backOffOnError`, and `throttle_on_success` at `ReadyForQuery`.

**Tests**

`tests/reject_throttle.js` — 21 cases: option plumbing, throttle-vs-herd, single prober, slow-start recovery and disengagement (including large backlogs and `max: 1`), the drop path staying intact via `max_lifetime`, and the boundaries (RST reject, connect_timeout, `off == vanilla`). Wired into `test:esm`; kept as a separate file for reviewability — happy to fold into `tests/index.js` if you'd prefer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "build:deno": "node transpile.deno.js",
     "build:cf": "node transpile.cf.js",
     "test": "npm run test:esm && npm run test:cjs && npm run test:deno",
-    "test:esm": "node tests/index.js",
+    "test:esm": "node tests/index.js && node tests/reject_throttle.js",
     "test:cjs": "npm run build:cjs && cd cjs/tests && node index.js && cd ../../",
     "test:deno": "npm run build:deno && cd deno/tests && deno run --no-lock --allow-all --unsafely-ignore-certificate-errors index.js && cd ../../",
     "lint": "eslint src && eslint tests",
--- a/src/connection.js
+++ b/src/connection.js
@@ -49,7 +49,7 @@ const errorFields = {
   82  : 'routine'            // R
 }
 
-function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose = noop } = {}) {
+function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose = noop, requeue = noop } = {}) {
   const {
     sslnegotiation,
     ssl,
@@ -362,6 +362,58 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     setTimeout(connect, closedTime ? Math.max(0, closedTime + delay - performance.now()) : 0)
   }
 
+  function backOffOnError() {
+    // Escalate the shared backoff on a real error — the increment closed() does on a
+    // drop. Callers recompute `delay` from it unconditionally (as the original does),
+    // so only this escalation is gated: called `hadError && ...` on the drop path, and
+    // unconditionally from our establishment reconnect (a close there is always an error).
+    options.shared.retries++
+  }
+
+  function reconnect_with_reject_throttle() {
+    const t = options.shared.throttle
+    if (t.prober === null || t.prober === connection)
+      prober_reconnect()
+    else
+      stand_down()
+  }
+
+  function prober_reconnect() {
+    // Become / stay the lone prober and pace the retry exactly as closed() paces a
+    // drop; every other establishing connection blocks behind us until we hand off.
+    // closedTime is set by closed() (as for a drop) — reconnect() only reads it.
+    options.shared.throttle.prober = connection
+    backOffOnError()
+    delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
+    reconnect()
+  }
+
+  function stand_down() {
+    // A non-prober open failed (a slow-start admit): collapse the ramp and give up
+    // this slot — hand the query back to the pool instead of retrying (which would
+    // re-form the herd). The prober keeps probing, now paced by the higher retries.
+    const t = options.shared.throttle
+    t.ramp = 1
+    t.allowance = 0
+    requeue(initial)
+    initial = null
+    onclose(connection, Errors.connection('CONNECTION_CLOSED', options, socket))
+  }
+
+  function throttle_on_success() {
+    // Only the prober's success clocks a recovery round: refill the admit allowance
+    // with the current batch and double the ramp (capped at the pool max). We do NOT
+    // null the prober here — it stays engaged (as this now-healthy connection) until
+    // throttleAdmit hands the baton to the next waiter, or to null once the backlog
+    // is drained. Admit successes are silent, else N concurrent successes compound it.
+    const t = options.shared.throttle
+    if (!options.reject_throttle || t.prober !== connection)
+      return
+
+    t.allowance += t.ramp
+    t.ramp = Math.min(t.ramp * 2, options.max)
+  }
+
   function connected() {
     try {
       statements = {}
@@ -447,12 +499,16 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     socket.removeAllListeners()
     socket = null
 
-    if (initial)
-      return reconnect()
+    if (initial) {
+      if (!options.reject_throttle)
+        return reconnect()
+      closedTime = performance.now()
+      return reconnect_with_reject_throttle()
+    }
 
     !hadError && (query || sent.length) && error(Errors.connection('CONNECTION_CLOSED', options, socket))
     closedTime = performance.now()
-    hadError && options.shared.retries++
+    hadError && backOffOnError()
     delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
     onclose(connection, Errors.connection('CONNECTION_CLOSED', options, socket))
   }
@@ -566,6 +622,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
 
       initial && !initial.reserve && execute(initial)
       options.shared.retries = retries = 0
+      throttle_on_success()
       initial = null
       return
     }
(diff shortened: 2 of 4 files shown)


fabio-vapi:
Hi @porsager, a gentle ping on this when you get a chance, no rush.

It's an opt-in `reject_throttle` option (default off, so behavior is identical to today) that fixes a connection establishment stampede against proxied/serverless Postgres: on a clean mid-handshake close, every pooled connection retries immediately and in lockstep with no backoff. #1179 has the full writeup of the root cause, a reproduction (baseline plus a deliberately catastrophic one), and the design rationale.

For motivation: we hit a production outage on Neon that came down to this. Admittedly our connection pool was oversized, which only exacerbated it, but the problem is fundamentally there regardless. Even a sensibly sized pool retries every connection at once with no backoff, so as soon as the endpoint starts dropping connections during establishment the whole pool stampedes it.

The PR ships a dedicated 21-case test suite, and it passes across every Node and Postgres version in the matrix (verified green on my fork; the checks here are just waiting on your workflow approval, since it's my first contribution). Happy to adjust the naming, design, or scope to whatever fits the project. Thanks for postgres.js!

PR #1198, opened 2026-08-21
Fix: Do not mutate parameters when serializing in Bind - fixes multiple serialization on retry

### LINKED ISSUE #1197: Parameters are serialized twice when a prepared query is retried - jsonb stored as string, boolean flips to false, bytea double-encoded (3.4.9)
### Summary

When a prepared query is retried after a `FetchPreparedStatement` or `RevalidateCachedQuery` error, its parameters go through the serializers a **second time**, and every parameter whose serializer is not idempotent is silently corrupted before being written to the database:

- `json`/`jsonb` values are `JSON.stringify`'d twice and stored as strings
- `boolean` `true` flips to `false` (`'t' === true` is `false`, so `'t'` re-serializes to `'f'`)
- `bytea` buffers turn into the hex encoding of their previous hex string

Reproduced deterministically on **postgres@3.4.9** (current `latest`) against PostgreSQL 16.15, Node v23.11.0.

### Mechanism

`Bind` (src/connection.js:959) stores each serialized value back into `query.parameters`:

```js
parameters[i] = x = type in options.serializers
  ? options.serializers[type](x)
  : '' + x
```

When the server answers with an error whose routine is in `retryRoutines` (src/connection.js:25-28), the query is retried on the same query object (src/connection.js:540-541 → `retry()` at src/connection.js:821), so `Bind` runs again - this time over the already-serialized values.

Both retry triggers are ordinary production events: `RevalidateCachedQuery` is the `cached plan must not change result type` error after a migration runs while an app keeps executing cached prepared statements, and `FetchPreparedStatement` fires when a prepared statement disappears server-side (e.g. `DISCARD ALL`/`DEALLOCATE` via a pooler). The retry behavior itself dates back to the v3 rewrite and exists to handle exactly these errors transparently (#120).

This also explains #678 - the non-deterministic "jsonb saved as an escaped string" corruption reported behind Supavisor/PgBouncer in transaction mode. A pooler discarding prepared statements server-side makes the `FetchPreparedStatement` retry fire routinely, and both workarounds found there fit this root cause: `prepare: false` avoids the retry path entirely (only `query.prepared` queries are retried), and an interleaved `.describe()` recreates the statement so the retry never happens.

### Reproduction

Single file. `deallocate all` triggers the `FetchPreparedStatement` retry; an `alter table` between two executions of the same query triggers `RevalidateCachedQuery` and reproduces it identically.

```js

const sql = postgres({ /* connection */, max: 1 })

await sql`create table retry_repro (config jsonb, enabled bool, data bytea)`

const insert = () => sql`insert into retry_repro values (${ sql.json({ foo: 'bar' }) }, ${ true }, ${ Buffer.from('hello') })`

await insert()
await sql`deallocate all`
await insert()

console.log(await sql`select jsonb_typeof(config) as config_type, enabled, data from retry_repro`)
await sql.end()
```

Output - first row was written before the retry, second row after it:

```
Result(2) [
  { config_type: 'object', enabled: true,  data: <Buffer 68 65 6c 6c 6f> },
  { config_type: 'string', enabled: false, data: <Buffer 5c 78 36 38 36 35 36 63 36 63 36 66> }
]
```

### Fix

Serializing into a local variable without mutating `query.parameters` fixes it; the full test suite still passes. I have a one-line fix plus tests covering both retry routines and all three affected built-in serializers - PR incoming.

Ne7Le4DeR:
Fixes #1197

`Bind` stored each serialized value back into `query.parameters`, so when a prepared query was retried after a `FetchPreparedStatement` or `RevalidateCachedQuery` error, the already serialized values went through the serializers a second time: `json`/`jsonb` objects got stored as JSON strings, `boolean true` silently flipped to `false`, and `bytea` buffers were hex-encoded twice. See #1197 for the full mechanism and a standalone reproduction.

Serialize into a local variable instead and leave `query.parameters` untouched - the retry itself is unchanged and still transparent, it just binds the original values now.

- The first commit adds six tests covering both retry routines and all three affected built-in serializers; each fails without the fix.
- The second commit is the one-line fix.
- The full test suite passes (ESM run against PostgreSQL 16).

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -956,7 +956,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
         return b.i32(0xFFFFFFFF)
 
       type = types[i]
-      parameters[i] = x = type in options.serializers
+      x = type in options.serializers
         ? options.serializers[type](x)
         : '' + x
 
--- a/tests/index.js
+++ b/tests/index.js
@@ -1824,6 +1824,87 @@ t('Recreate prepared statements on RevalidateCachedQuery error', async() => {
   ]
 })
 
+t('Does not re-serialize json parameters when retrying on RevalidateCachedQuery error', async() => {
+  const update = () => sql`update test set config = ${ sql.json({ warmup: 2100000000, ticksLimit: 5000 }) } returning *`
+  await sql`create table test (id int, config jsonb)`
+  await sql`insert into test values (1, '{}')`
+  await update()
+  await sql`alter table test add column extra int`
+  await update()
+  return [
+    'object',
+    (await sql`select jsonb_typeof(config) as type from test`)[0].type,
+    await sql`drop table test`
+  ]
+})
+
+t('Does not re-serialize json parameters when retrying on FetchPreparedStatement error', async() => {
+  const insert = () => sql`insert into test (config) values (${ sql.json({ warmup: 2100000000 }) })`
+  await sql`create table test (config jsonb)`
+  await insert()
+  await sql`deallocate all`
+  await insert()
+  return [
+    'object,object',
+    (await sql`select jsonb_typeof(config) as type from test`).map(x => x.type).join(),
+    await sql`drop table test`
+  ]
+})
+
+t('Does not re-serialize boolean parameters when retrying on RevalidateCachedQuery error', async() => {
+  const update = () => sql`update test set enabled = ${ true } returning *`
+  await sql`create table test (id int, enabled bool)`
+  await sql`insert into test values (1, false)`
+  await update()
+  await sql`alter table test add column extra int`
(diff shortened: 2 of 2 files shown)


Ne7Le4DeR:
Hi @porsager, sorry for the ping. This one is a silent data corruption bug, and a pretty nasty one - no error is thrown, wrong values just end up in the table.

The fix is one line plus tests, details are in the description. Would appreciate a look when you have a minute.

porsager:
Thank you for the ping @Ne7Le4DeR - I'll look into it! My time is unfortunately limited and I've had no way of keeping up with the inflow of AI wall of text issues.

porsager:
Great PR and catch! Thank you!

PR #1211, opened 2026-09-02
Run onexecute regardless of max_pipeline so begin() works at 0

### LINKED ISSUE #1210: sql.begin() always throws UNSAFE_TRANSACTION when max_pipeline is 0
**Version:** 3.4.9 (also current `master`, 411429e)
**Node:** v26.8.1, plain PostgreSQL 16 (no pooler involved in the repro)

With the connection option `max_pipeline: 0`, every `sql.begin()` rejects with `UNSAFE_TRANSACTION: Only use sql.begin, sql.reserved or max: 1`, even though the transaction is being run through `sql.begin`.

### Reproduction

```js
const url = process.env.DATABASE_URL
for (const [label, opts] of [['default', {}], ['max_pipeline: 1', { max_pipeline: 1 }], ['max_pipeline: 0', { max_pipeline: 0 }]]) {
  const sql = postgres(url, { max: 4, ...opts })
  try { const r = await sql.begin(tx => tx`select 1 as ok`); console.log(label, 'OK', r) }
  catch (e) { console.log(label, 'FAIL', e.code, e.message) }
  finally { await sql.end({ timeout: 2 }) }
}
```

Output on 3.4.9 / master:

```
default OK Result(1) [ { ok: 1 } ]
max_pipeline: 1 OK Result(1) [ { ok: 1 } ]
max_pipeline: 0 FAIL UNSAFE_TRANSACTION UNSAFE_TRANSACTION: Only use sql.begin, sql.reserved or max: 1
```

### Cause

`execute(q)` in `src/connection.js` returns a single `&&` chain:

```js
      build(q)
      return write(toBuffer(q))
        && !q.describeFirst
        && !q.cursorFn
        && sent.length < max_pipeline
        && (!q.options.onexecute || q.options.onexecute(connection))
```

`begin()` in `src/index.js` sends `BEGIN` via `sql.unsafe('begin ...', [], { onexecute })` and relies on that `onexecute` hook to capture the connection and `move(c, reserved)` / set `c.reserved`. Because the hook sits after `sent.length < max_pipeline` in the same chain, `max_pipeline: 0` (`0 < 0` is false) short-circuits before it, so the hook never runs. The connection is never reserved, and when the BEGIN's CommandComplete arrives the guard

```js
    if (result.command === 'BEGIN' && max !== 1 && !connection.reserved)
      return errored(Errors.generic('UNSAFE_TRANSACTION', 'Only use sql.begin, sql.reserved or max: 1'))
```

fires. One expression is doing two unrelated jobs: "may the pool pipeline another query onto this connection" and "may the reservation hook run".

### Why `max_pipeline: 0`

Behind a transaction-mode connection pooler (e.g. Supavisor/PgBouncer) a second query pipelined onto the socket before the first is answered is not handled correctly, so disabling pipelining entirely is the natural setting there. `max_pipeline: 1` still lets one extra query be queued on the socket.

### Expected / actual

- Expected: `sql.begin()` works with `max_pipeline: 0` exactly as with any other value; the option should only affect how many queries are queued on a connection.
- Actual: every `sql.begin()` rejects with `UNSAFE_TRANSACTION`.

A PR with a fix and tests follows.

reinierlakhan:
Fixes #1210

### Cause

`execute(q)` in `src/connection.js` returns one `&&` chain that does two unrelated jobs: it decides whether the pool may pipeline another query onto this connection, **and** it runs the query's `onexecute` hook as its last term:

```js
      return write(toBuffer(q))
        && !q.describeFirst
        && !q.cursorFn
        && sent.length < max_pipeline
        && (!q.options.onexecute || q.options.onexecute(connection))
```

`begin()` sends `BEGIN` with `{ onexecute }` and relies on that hook to capture the connection, `move(c, reserved)` and set `c.reserved`. With `max_pipeline: 0` the `sent.length < max_pipeline` term is always false, so the chain short-circuits before the hook. The connection is never reserved, and the `UNSAFE_TRANSACTION` guard in `CommandComplete` rejects every transaction.

### Change

Run the hook whenever the query was actually written, independent of pipeline capacity, and keep the return value's meaning ("may the pool pipeline more onto this connection") unchanged:

```js
      const written = write(toBuffer(q))
      written && q.options.onexecute && q.options.onexecute(connection)
      return written
        && !q.describeFirst
        && !q.cursorFn
        && sent.length < max_pipeline
```

**`max_pipeline > 0`: behaviour is identical.** `onexecute` in `begin()` (the only caller) returns a truthy value (the assigned `c.reserved` function), so the old chain was truthy exactly when the new one is, and the hook ran in exactly the cases it runs now (previously it also only ran when `write()` returned truthy, which `written &&` preserves).

**`max_pipeline: 0`:** `execute()` reserves the connection and returns `false`, so `go()` moves it to `full`. `c.reserved` is set, so the BEGIN passes the guard. On the BEGIN's ReadyForQuery (status `T`), `connection.reserved()` runs, finds the transaction's queue empty and moves the connection back to `reserved`. The transaction callback only starts after that (it is awaited behind the BEGIN), and each statement it issues goes through `begin()`'s `handler`: `execute()` returns `false` -> `move(c, full)`; anything issued while in `full` is pushed onto the transaction's own `queries` and drained one at a time by `c.reserved()` on each ReadyForQuery. Ordering is preserved and nothing is pipelined onto the socket, which is the point of the setting.

### Tests

Three tests added to `tests/index.js`, each on a client with `{ max: 4, max_pipeline: 0 }`:

- a single-statement `sql.begin(sql => sql\`select 1 as x\`)` returns `1`
- two sequential statements in one transaction (`set_config(..., true)` then `current_setting`, so it also proves they ran on the same transaction)
- two concurrent statements in one transaction (array form), which exercises the `queries` queue path

Run in isolation against a plain PostgreSQL 16 through the repo's `tests/test.js` harness: all three pass with this change; without it the first fails with `UNSAFE_TRANSACTION`. `eslint src tests` with the repo config is clean.

**The full suite did not run here**: `tests/bootstrap.js` needs `psql`/`createdb`/`dropdb` on PATH and a server it may reconfigure (`ssl=on`, `wal_level=logical`, prepared transactions), which this environment does not have. CI should cover it.

I left `cjs/`, `deno/` and `cf/` untouched, following the pattern of separate `build` commits in this repo; happy to regenerate them in this PR if preferred.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -170,11 +170,12 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
         : (query = q, query.active = true)
 
       build(q)
-      return write(toBuffer(q))
+      const written = write(toBuffer(q))
+      written && q.options.onexecute && q.options.onexecute(connection)
+      return written
         && !q.describeFirst
         && !q.cursorFn
         && sent.length < max_pipeline
-        && (!q.options.onexecute || q.options.onexecute(connection))
     } catch (error) {
       sent.length === 0 && write(Sync)
       errored(error)
--- a/tests/index.js
+++ b/tests/index.js
@@ -332,6 +332,27 @@ t('Helpers in Transaction', async() => {
   ))[0].x]
 })
 
+t('Transaction works with max_pipeline 0', async() => {
+  const sql = postgres({ ...options, max: 4, max_pipeline: 0 })
+  return [1, (await sql.begin(sql => sql`select 1 as x`))[0].x, await sql.end()]
+})
+
+t('Transaction with sequential queries works with max_pipeline 0', async() => {
+  const sql = postgres({ ...options, max: 4, max_pipeline: 0 })
+  return ['testing', await sql.begin(async sql => {
+    await sql`select set_config('postgres_js.test', 'testing', true)`
+    return (await sql`select current_setting('postgres_js.test') as x`)[0].x
+  }), await sql.end()]
+})
+
+t('Transaction with concurrent queries works with max_pipeline 0', async() => {
+  const sql = postgres({ ...options, max: 4, max_pipeline: 0 })
+  return ['testing', (await sql.begin(sql => [
+    sql`select set_config('postgres_js.test', 'testing', true)`,
+    sql`select current_setting('postgres_js.test') as x`
+  ]))[1][0].x, await sql.end()]
+})
+
 t('Undefined values throws', async() => {
   let error
 


v0idpwn:
@reinierlakhan I had written a similar fix, just PRed it on #1218. I believe there's a regression in your branch that causes another query to end up in the connection that should be reserved for the transaction. I wrote a test reproducing it [here](https://github.com/porsager/postgres/pull/1218/changes#diff-9c265283de2ab902547ac9c10806cee40c019f48d842ac7e8e014f868ca8cf9cR318)

Ruckus000:
With the default `max_pipeline` (and with `1`), this change lets an unrelated query run **inside another caller's transaction**.

The PR description says `begin()`'s `onexecute` "returns a truthy value (the assigned `c.reserved` function)". It doesn't: `onexecute` in `src/index.js` has a block body with no `return`, so it returns `undefined`. On `master`, that `undefined` made `execute()` return falsy for BEGIN, and `go()` moved the connection to `full`. With this PR, `execute()` returns `written && ... && sent.length < max_pipeline`, which is `true` for BEGIN. `go()` then moves the freshly reserved connection back to `busy`, and the next ordinary query is pipelined onto it while BEGIN is open.

Repro: plain PostgreSQL 16, `max: 1`. The outside query is issued only after BEGIN has been written to the socket:

```js

for (const max_pipeline of [undefined, 1, 0]) {
  let beginWritten
  const begun = new Promise(r => (beginWritten = r))
  const sql = postgres(process.env.DATABASE_URL, {
    max: 1,
    ...(max_pipeline === undefined ? {} : { max_pipeline }),
    debug: (_, q) => /^begin/i.test(q.trim()) && beginWritten()
  })
  await sql`select 1`
  const tx = sql.begin(async t => (await t`select now()::text as ts, pg_sleep(0.3)`)[0].ts)
  await begun
  const outside = sql`select now()::text as ts`
  const [txTs, [row]] = await Promise.all([tx, outside])
  console.log(`max_pipeline=${max_pipeline ?? 'default'}: outside query ran inside the transaction = ${row.ts === txTs}`)
  await sql.end()
}
```

`now()` is the transaction start time, so equal values mean the outside query ran inside the other transaction.

| `src/connection.js` `execute()` | default | `1` | `0` |
|---|---|---|---|
| 3.4.9 / master | false | false | throws (`Cannot set properties of undefined (setting 'onclose')`) |
| this PR (#1211) | **true** | **true** | false |
| #1218 | false | false | false |
| always call `onexecute`, return `false` when present | false | false | false |

The #1211 result reproduced 3 out of 3 times. #1218 keeps `onexecute`'s falsy return in the chain, which is what keeps the connection out of `busy` until BEGIN completes, so it doesn't have this problem. I'd suggest #1218 over this one, or keeping the return falsy whenever `onexecute` is present.

PR #1214, opened 2026-09-08
Bound reconnects after clean startup closes

### LINKED ISSUE #1193: Clean socket close with a pending initial query never settles it and spins an unbounded zero-delay reconnect loop
### Summary

A socket that closes **cleanly** (FIN, `hadError === false`, no prior `ErrorResponse`) while a query is parked as a connection's `initial` leaves that query pending **forever**, and turns `reconnect()` into a **zero-delay reconnect loop** that `connect_timeout` and `max_lifetime` cannot bound. Affects 3.4.9 and current master.

### Mechanism (`src/connection.js`, `closed()`, lines 436-458 on 3.4.9)

```js
idleTimer.cancel()
lifeTimer.cancel()
connectTimer.cancel()        // cancelled on EVERY close ...

socket.removeAllListeners()
socket = null

if (initial)
  return reconnect()         // returns BEFORE every settle path

!hadError && (query || sent.length) && error(...)
closedTime = performance.now()   // never reached on that branch
```

Three consequences, each verified against the source:

1. **The parked query never settles.** A clean close raises no `'error'` event and carries no `ErrorResponse`, so nothing has rejected `initial` before `closed()` runs. The `if (initial) return reconnect()` branch returns before `error(...)`/`errored(...)` and before `onclose(...)`, so the query is in no queue and on no timer that can ever settle it. (An RST rejects via the `'error'` event; a server `ErrorResponse` rejects via `errored()`; only the clean FIN is unhandled.)
2. **`closedTime` never advances**, so `reconnect()` at line 361-363 (`setTimeout(connect, closedTime ? Math.max(0, closedTime + delay - performance.now()) : 0)`) fires with **delay 0** on every iteration.
3. **No configured bound can fire.** `connectTimer.cancel()` runs on every close and `connect()` restarts it (line 343); `lifeTimer` the same (lines 444 / 371). With a sub-millisecond close/reconnect cycle neither ever elapses.

### Reproduction (standalone, no Postgres needed)

```js
import net from 'node:net'

let attempts = 0
const server = net.createServer(s => { attempts++; s.end() }) // accept, clean FIN
await new Promise(r => server.listen(0, '127.0.0.1', r))

const sql = postgres(`postgres://u:p@127.0.0.1:${server.address().port}/db`, {
  max: 3, idle_timeout: 20, connect_timeout: 10, prepare: false,
})

const q = sql`select 1`
setTimeout(() => console.log({ attempts }), 8000) // pending, ~5,000+ attempts
await q // never settles
```

Measured on 3.4.9 / Node 22: the query is still pending at 8 seconds with **5,080 connection attempts** (~635 attempts/second, one CPU-bound reconnect loop per parked connection). Once `max` connections are looping, every further query lands in `queries` with no `onopen` or `onclose` ever coming, so the whole pool stalls.

Real-world trigger: a serverless function whose upstream pooler (Supavisor) closed connections with a clean FIN during a pooler-side event; the pending query rode all the way to the platform's 300-second function ceiling instead of failing at `connect_timeout: 10`. We measured the pooler's close verb at the socket level: every close it initiates is FIN + `hadError=false` (an over-limit refusal is preceded by an `ErrorResponse`, which settles correctly; the FIN itself never does).

### Suggested fix

In `closed()`, treat a close-with-`initial` like every other close: advance `closedTime`, ride the shared backoff, and cap consecutive attempts, settling the parked query at the cap:

```js
if (initial) {
  closedTime = performance.now()
  options.shared.retries++
  delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
  if (++closeRetries < 5)          // closeRetries: new per-connection counter,
    return reconnect()             // reset alongside `retries` on ReadyForQuery
  closeRetries = 0
  errored(Errors.connection('CONNECTION_CLOSED', options, socket))
  return onclose(connection, Errors.connection('CONNECTION_CLOSED', options, socket))
}
```

A transient close still reconnects and recovers (we pin a peer that FINs two handshakes then serves honestly: the parked query resolves), while a peer that closes every attempt rejects the query with `CONNECTION_CLOSED` after five paced attempts (~1s with the default backoff) instead of hanging forever. Verified against the repro above: rejected in ~800ms after exactly 5 attempts.

We are carrying this as a pnpm patch in production (all three shipped builds) and are happy to open a PR if the approach looks right.

Related: #1192 (same `closed()`/`ReadyForQuery` neighbourhood, different defect: the discarded `fetchArrayTypes()` promise).

diasvtumane:
If a server sends FIN during startup, the first query can hang forever while we reconnect with no delay.

Use the shared backoff for these retries and reject with `CONNECTION_CLOSED` if the next attempt would exceed a `connect_timeout` budget (30 seconds when disabled). Clear the budget after recovery or an error so later queries get a fresh start. Reserved connections also reset the retry counter after startup.

Added seven TCP regression tests for repeated closes, recovery, queued queries and reserves, and backoff beyond the timeout.

Fixes #1193.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -87,6 +87,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     , statementId = Math.random().toString(36).slice(2)
     , statementCount = 1
     , closedTime = 0
+    , closeRunStart = 0
     , remaining = 0
     , hostIndex = 0
     , retries = 0
@@ -447,13 +448,21 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     socket.removeAllListeners()
     socket = null
 
-    if (initial)
-      return reconnect()
-
-    !hadError && (query || sent.length) && error(Errors.connection('CONNECTION_CLOSED', options, socket))
     closedTime = performance.now()
-    hadError && options.shared.retries++
+    if (hadError || initial)
+      options.shared.retries++
     delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
+
+    if (initial) {
+      closeRunStart || (closeRunStart = closedTime)
+      // Do not schedule a retry beyond the connection's clean-close budget.
+      if (closedTime + delay <= closeRunStart + (options.connect_timeout || 30) * 1000)
+        return reconnect()
+      errored(Errors.connection('CONNECTION_CLOSED', options, socket))
+    }
+
+    closeRunStart = 0
+    !hadError && (query || sent.length) && error(Errors.connection('CONNECTION_CLOSED', options, socket))
     onclose(connection, Errors.connection('CONNECTION_CLOSED', options, socket))
   }
 
@@ -560,12 +569,12 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
       }
 
       if (needsTypes) {
-        initial.reserve && (initial = null)
+        initial.reserve && (options.shared.retries = retries = closeRunStart = 0, initial = null)
         return fetchArrayTypes()
       }
 
       initial && !initial.reserve && execute(initial)
-      options.shared.retries = retries = 0
+      options.shared.retries = retries = closeRunStart = 0
       initial = null
       return
     }
--- a/tests/connection.js
+++ b/tests/connection.js
@@ -1,0 +1,186 @@
+import assert from 'assert'
+import { createServer } from 'net'
+import postgres from '../src/index.js'
+
+const delay = ms => new Promise(resolve => setTimeout(resolve, ms))
+    , ready = message('Z', [73])
+
+export function cleanClose() {
+  let attempts = 0
+    , closes = 0
+  const retries = []
+
+  return peer(() => (++attempts, 'close'), {
+    connect_timeout: 0.15,
+    backoff: retry => (retries.push(retry), 0.025),
+    onclose: () => closes++
+  }, async sql => {
+    const error = await sql`select 1`.catch(error => error)
+    assert.strictEqual(error.code, 'CONNECTION_CLOSED')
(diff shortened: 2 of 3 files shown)


diasvtumane:
Hey @porsager , so sorry for the ping. This one fixes a month old issue. We hit this issue in production and it took me a minute to figure out what exactly is wrong. I will patch it in our own repo but could you take a look when you get a chance?

jeffcoughlin:
Thanks for investigating this. I may be encountering the same failure mode in a Next.js 16.3.4 application using Drizzle, postgres.js 3.4.9, and Supabase/Supavisor.

Intermittently, all database-backed routes begin waiting for tens of seconds or several minutes, while non-database routes and the dev server can remain responsive. Restarting the Node/Next.js process immediately restores normal database behavior.

I have observed requests remaining pending for more than nine minutes even though the postgres.js client is configured with:

- `connect_timeout: 5`
- `idle_timeout: 20`
- `max: 10`
- `prepare: false`
- PostgreSQL `statement_timeout: 25000`

Some requests report PostgreSQL error `57014` (`canceling statement due to statement timeout`), while others remain pending far beyond either configured timeout. Delayed database requests sometimes complete together or only when the development process is stopped.

I have not applied this branch yet, so I cannot confirm that #1214 resolves my case. However, the clean-FIN/Supavisor startup behavior described in #1193, along with the queued-query symptoms, appears closely aligned with what I am seeing.

Please let me know if there are specific diagnostics I can capture or if testing this patch against my application would be helpful. I would also be interested to know whether this is likely to be included in an upcoming release.

fabio-vapi:
Actually I believe this PR of mine does a better job at controlling the herd effect of retrying connections, let me know what you think.

https://github.com/porsager/postgres/issues/1179

Issue #1204, opened 2026-08-30
Connection of one in-process transaction being handed to a separate transaction, causing corruption


darkgnotic:
**Version:** `postgres@3.4.7` · Node 24 · PostgreSQL 16 (AWS Aurora)

---

## Summary

We observed one connection being used by two `sql.begin()` transactions,
and that second transaction's `COMMIT` then **committed the first transaction's partial, unfinished writes**

## What we run

One pool (`max: 5`) shared by several concurrent `sql.begin()` workloads:

- a long-lived **insert**  transaction that pipelines a few thousand `INSERT`s,
   then a single final `UPDATE` marker, all in one `sql.begin`;
- a concurrent short **purge** transaction (a `DELETE` within `sql.begin`);
- occasional table scans done in transaction snapshots

Options: `max_lifetime` 300–600s (randomized), `idle_timeout: 60`, TLS on, over a network
where connections can half-close. (We can't say whether the pool was saturated at the moment
of the collision; we only know two specific transactions, the writer and the purge, ended up
 sharing one connection.)

## Evidence (from the PostgreSQL server log)

Backend `74671` was in the middle of an **insert** transaction: 

```
19:00:28 [74671]:LOG:  AUDIT: SESSION,1366,1,WRITE,INSERT,,,"INSERT INTO "
19:00:29 [74671]:LOG:  AUDIT: SESSION,1367,1,WRITE,INSERT,,,"INSERT INTO "
```

Then on that same backend:

```
19:00:29  [74671] WARNING:  there is already a transaction in progress   (BeginTransactionBlock)
```

Corresponding exactly to server log lines indicating the purger was executing its transaction:

```
2026-08-29T19:00:29.52707903Z  | INFO | Purging PG changes before 724at0ueo ...
2026-08-29T19:00:29.528707661Z | WARN | {"level":"WARN","pid":27,"worker":"change-streamer","workerIndex":0,"severity":"WARNING","code":"25001","message":"pg notice","routine":"BeginTransactionBlock"}
2026-08-29T19:00:29.616657299Z | INFO | Purged 6334 PG changes before 724at0ueo (89.49 ms)
```
(The `DELETE` statements from the purger are wrapped in a CTE and thus is filtered from audit logs as "reads")

After that `WARNING`, backend `74671` logged **nothing further except a clean disconnection**:

```
19:01:30  [74671] LOG:  disconnection: session time: 0:05:11.707 ... user=... database=...
```

[...]

Issue #1133, opened 2025-11-24
Uncatchable exception on idle in transcation timeout


k0nserv:
Hello!

If you trigger a connection termination due to idle in transaction timeout `postgres` throws an error that is not catchable from the callsite.

**Minimal repro:**

```javascript
const postgres = require("postgres");

const sql = postgres({
  host: process.env.PGHOST || "localhost",
  port: process.env.PGPORT ? Number(process.env.PGPORT) : 5432,
  database: process.env.PGDATABASE || "postgres_connection_issue",
  user: process.env.PGUSER,
});

const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));

async function main() {
  console.log("\n--- minimal idle timeout repro ---");
  try {
    await sql.begin(async (tx) => {
      await tx`SET idle_in_transaction_session_timeout = 50`;
      await sleep(200);
    });
  } catch (err) {
    // The driver throws a normal connection-closed error after the server kills
    // the backend for idling in a transaction. We log that and keep going.
    console.error("caught error:", err?.message || err);
  }
}

process.on("uncaughtException", (err) => {
  console.error("uncaught exception:", err);
});

main();
```

The stack trace is

```
    at Immediate.nextWrite ([..]/node_modules/postgres/cjs/src/connection.js:250:22)
    at process.processImmediate (node:internal/timers:483:21)
```

This seems to be a race condition between `nextWrite` and the socket being closed.

This becomes a problem because the exception cannot be caught without using `process.on`, ideally this exception would not be raised.

PR #1215, opened 2026-09-09
Reject transaction work after its connection closes

### LINKED ISSUE #1216: Disconnected transaction handles can read another user's rows through RLS
A `sql.begin()` transaction handle can keep issuing queries after its backend disconnects. Once the pool reconnects and starts another transaction, the old handle executes against that replacement transaction.

This has security implications for applications that use a transaction-local setting to identify users in RLS policies. I reproduced the original user's transaction handle reading the replacement user's row with RLS enabled and a non-owner database role. PostgreSQL evaluates the policy using the replacement transaction's `app.user_id`, so the stale query inherits that user's access.

The fix is in #1215. The related connection-reuse report in #1204 does not establish this specific RLS case, and I haven't confirmed that its production incident has the same cause.

### Reproduction

Tested with Node 22.22.0, PostgreSQL 17.11, and upstream commit `411429e7bd7a3d61155ca9a70a97c111823702ea` (version 3.4.9).

Save this as `repro-rls.mjs` in a checkout of that commit. The script creates two synthetic users' rows and a separate login with only SELECT access to the test table. It verifies that RLS is active and each live transaction initially sees only its own row.

```js
import assert from 'node:assert/strict'
import postgres from './src/index.js'

const role = 'postgres_js_rls_test'
const owners = { original: 'original', replacement: 'replacement' }
const rows = Object.values(owners).map(user_id => ({ user_id, secret: user_id + ' secret' }))
const admin = postgres({ max: 1 })
const pool = postgres({ user: role, database: 'postgres', max: 1, fetch_types: false })
let finish
  , ready
const gate = new Promise(resolve => finish = resolve)
const connected = new Promise(resolve => ready = resolve)

try {
  await admin`create role ${ admin(role) } login`
  await admin`create table rls_repro (user_id text not null, secret text not null)`
  await admin`insert into rls_repro ${ admin(rows) }`
  await admin`alter table rls_repro enable row level security`
  await admin`create policy owner_only on rls_repro
    using (user_id = current_setting('app.user_id', true))`
  await admin`grant select on rls_repro to ${ admin(role) }`

  const failed = pool.begin(async sql => {
    await sql`select set_config('app.user_id', ${ owners.original }, true)`
    const [{ active }] = await sql`select row_security_active('rls_repro') as active`
    assert.equal(active, true)
    assert.deepEqual(Array.from(await sql`select * from rls_repro`), [rows[0]])
    const [{ pid }] = await sql`select pg_backend_pid() as pid`
    ready({ disconnected: sql, pid })
    await gate
  }).catch(error => error)

  const { disconnected, pid } = await Promise.race([
    connected,
    failed.then(error => { throw error })
  ])
  await admin`select pg_terminate_backend(${ pid }::int)`
  assert.equal((await failed).code, 'CONNECTION_CLOSED')

  await pool.begin(async sql => {
    await sql`select set_config('app.user_id', ${ owners.replacement }, true)`
    const [{ active }] = await sql`select row_security_active('rls_repro') as active`
    assert.equal(active, true)
    assert.deepEqual(Array.from(await sql`select * from rls_repro`), [rows[1]])
    const result = await disconnected`select * from rls_repro`.catch(error => error)
    console.log(result)
    assert.equal(result.code, 'CONNECTION_CLOSED')
  })
} finally {
  finish()
  await new Promise(resolve => setImmediate(resolve))
  await pool.end({ timeout: 0 })
  await admin`drop table if exists rls_repro`
  await admin`drop role if exists ${ admin(role) }`
  await admin.end({ timeout: 0 })
}
```

Run it against a disposable PostgreSQL instance:

```sh
docker run --rm -d --name postgres-js-rls-repro -p 127.0.0.1:55432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgres:17
docker exec postgres-js-rls-repro pg_isready -U postgres
PGHOST=127.0.0.1 PGPORT=55432 PGUSER=postgres PGDATABASE=postgres node repro-rls.mjs
docker stop postgres-js-rls-repro
```

Wait for `pg_isready` to report that the server is accepting connections before running the script.

On upstream, the query through the disconnected transaction prints:

```text
Result(1) [ { user_id: 'replacement', secret: 'replacement secret' } ]
```

The assertion then fails. The expected result is a `CONNECTION_CLOSED` error: the old transaction should no longer be usable.

Running the same script against commit `18cffe016927e01481b9e746587da789f4260a4c` from #1215 returns `CONNECTION_CLOSED` and exits successfully. I verified both outcomes.

### Conditions and impact

The original callback must remain pending across a backend disconnect, the pool must reuse its connection object, and work must issue a query through the old handle after the replacement transaction establishes its user context. The reproduction uses `pg_terminate_backend` to force that sequence. It does not establish that an untrusted client can trigger it in every application.

The demonstrated impact is a read across the application's user boundary. The PR also includes regressions for an old callback committing or rolling back a replacement transaction, and for queued queries remaining pending after disconnect.

@porsager, could you assess this as a security issue and review #1215? It includes the fix, four regression tests, and before/after testing instructions.

ecsbeats:
Fixes #1216.

If a backend disconnects while a `sql.begin()` callback is awaiting other work, `begin()` rejects but the callback keeps running. Once the pool reconnects, that callback can issue queries on the replacement connection. Returning or throwing from it can also commit or roll back a different transaction. Queries still queued inside the original transaction can remain pending indefinitely.

Keep the close error on the transaction, reject its queued queries, and reject subsequent queries through that transaction's handler, including automatic commit/rollback. Clear the connection's pending write and query/result/error state before reuse, too. Without that reset, the first query after reconnect can receive the old backend's fatal error.

The four regression tests cover queued queries, a query through the disconnected transaction, and late commit/rollback. The latter three share a fixture that pauses the original callback until a replacement transaction is open. No new dependencies or public API changes.

Related to #1204. This reproduction specifically covers a backend disconnect followed by connection reuse; I haven't verified that it explains the original report.

### Reproduce

Save the following as `repro.mjs` in the repository root. It needs Node 22 and a PostgreSQL login that can terminate its own other sessions.

```js
import assert from 'node:assert/strict'
import postgres from './src/index.js'

const pool = postgres({ max: 1 })
const admin = postgres({ max: 1 })
let resume
  , connected
const gate = new Promise(resolve => { resume = resolve })
const ready = new Promise(resolve => { connected = resolve })
const failed = pool.begin(async sql => {
  connected((await sql`select 1`).state.pid)
  await gate
}).catch(error => error)

try {
  await admin`select pg_terminate_backend(${ await ready }::int)`
  assert.equal((await failed).code, 'CONNECTION_CLOSED')
  await pool.begin(async sql => {
    const [{ id: before }] = await sql`select txid_current()::text as id`
    resume()
    await new Promise(resolve => setImmediate(resolve))
    const [{ id: after }] = await sql`select txid_current()::text as id`
    console.log({ before, after })
    assert.equal(after, before)
  })
} finally {
  resume()
  await Promise.all([pool.end({ timeout: 0 }), admin.end({ timeout: 0 })])
}
```

For a disposable local server:

```sh
docker run --rm -d --name postgres-disconnect-repro \
  -p 127.0.0.1:55432:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust postgres:17
docker exec postgres-disconnect-repro pg_isready -U postgres
PGHOST=127.0.0.1 PGPORT=55432 PGUSER=postgres PGDATABASE=postgres node repro.mjs
docker stop postgres-disconnect-repro
```

Wait for `pg_isready` to report that the server is accepting connections before running the script. On upstream `411429e`, the assertion fails because the transaction IDs differ: the old callback has committed the replacement transaction. On this branch they stay equal and the script exits successfully. I verified both outcomes.

### Tests

The first commit, `0282301`, adds only the regression tests; the second applies the fix. With the PostgreSQL setup from [the existing test workflow](https://github.com/porsager/postgres/blob/411429e7bd7a3d61155ca9a70a97c111823702ea/.github/workflows/test.yml) in place:

```sh
git worktree add ../postgres-disconnect-before 0282301
(cd ../postgres-disconnect-before && PGUSER=postgres PGSOCKET=/var/run/postgresql npm run test:esm)
PGUSER=postgres PGSOCKET=/var/run/postgresql npm test
npm exec --yes --package=eslint@8.57.1 -- eslint src tests
```

The first run times out in `Disconnect rejects queued transaction queries and allows reconnect`. The fixed branch passes. The full suite needs both PostgreSQL servers from the workflow (ports 5432 and 5433), SSL, logical replication, prepared transactions, the PostgreSQL CLI tools, and Deno 1.x. Use disposable databases: the existing bootstrap changes server settings and recreates its test database and roles.

Validated against PostgreSQL 17.11:

- Node 12.22.12 and 22.22.0: all 268 tests pass in both ESM and CommonJS.
- Deno 1.46.3: all 268 tests pass.
- Bun 1.4.2: the standalone reproduction and focused probes for all four failure cases pass. The full suite cannot register tests because `tests/test.js:20` assumes a Node stack format; the same failure occurs on unmodified upstream.
- ESLint passes.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -438,6 +438,7 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
     remaining = 0
     incomings = null
     clearImmediate(nextWriteTimer)
+    chunk = nextWriteTimer = null
     socket.removeListener('data', data)
     socket.removeListener('connect', connected)
     idleTimer.cancel()
@@ -451,6 +452,9 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
       return reconnect()
 
     !hadError && (query || sent.length) && error(Errors.connection('CONNECTION_CLOSED', options, socket))
+    query = results = errorResponse = null
+    result = new Result()
+    rows = 0
     closedTime = performance.now()
     hadError && options.shared.retries++
     delay = (typeof backoff === 'function' ? backoff(options.shared.retries) : backoff) * 1000
--- a/src/index.js
+++ b/src/index.js
@@ -236,13 +236,19 @@ function Postgres(a, b) {
     const queries = Queue()
     let savepoints = 0
       , connection
+      , closedError
       , prepare = null
 
     try {
       await sql.unsafe('begin ' + options.replace(/[^a-z ]/ig, ''), [], { onexecute }).execute()
       return await Promise.race([
         scope(connection, fn),
-        new Promise((_, reject) => connection.onclose = reject)
+        new Promise((_, reject) => connection.onclose = error => {
+          closedError = error
+          while (queries.length)
+            queries.shift().reject(error)
+          reject(error)
+        })
       ])
     } catch (error) {
       throw error
@@ -290,6 +296,8 @@ function Postgres(a, b) {
 
       function handler(q) {
         q.catch(e => uncaughtError || (uncaughtError = e))
+        if (closedError)
+          return q.reject(closedError)
         c.queue === full
           ? queries.push(q)
           : c.execute(q) || move(c, full)
--- a/tests/index.js
+++ b/tests/index.js
@@ -2501,6 +2501,81 @@ t('Ensure transactions throw if connection is closed dwhile there is no query', 
   return ['CONNECTION_CLOSED', x.code]
 })
 
+t('Disconnect rejects queued transaction queries and allows reconnect', async() => {
+  const sql = postgres({ ...options, max_pipeline: 1, fetch_types: false })
+  let queries
+
+  try {
+    const error = await sql.begin(sql => {
+      queries = [
+        sql`select pg_terminate_backend(pg_backend_pid())`.execute(),
+        sql`select 1`.execute(),
+        sql`select 2`.execute()
+      ]
+      return Promise.allSettled(queries)
+    }).catch(x => x)
+
+    const results = await Promise.allSettled(queries)
+    const [{ x }] = await sql`select 1 as x`
+    return [
+      'CONNECTION_CLOSED,rejected,rejected,rejected,1',
+      [error.code, ...results.map(x => x.status), x].join(',')
+    ]
+  } finally {
+    await sql.end({ timeout: 0 })
+  }
+})
+
+t('Disconnected transaction cannot query a reused connection', async() =>
+  withDisconnectedTransaction(({ sql, disconnected }) => sql.begin(async sql => {
+    await sql`select set_config('postgres_js.test', 'replacement', true)`
+    const result = await disconnected`select current_setting('postgres_js.test') as x`.catch(x => x)
+    return ['CONNECTION_CLOSED', result.code]
+  }))
+)
+
+t('Disconnected transaction cannot commit a reused connection', async() =>
+  finishDisconnectedTransaction()
+)
+
+t('Disconnected transaction cannot roll back a reused connection', async() =>
+  finishDisconnectedTransaction(new Error('original callback failed'))
+)
+
+function finishDisconnectedTransaction(error) {
+  return withDisconnectedTransaction(({ sql, finish }) => sql.begin(async sql => {
+    const [{ x: before }] = await sql`select txid_current()::text as x`
+    finish(error)
+    await new Promise(resolve => setImmediate(resolve))
+    const [{ x: after }] = await sql`select txid_current()::text as x`
+    return [before, after]
+  }))
+}
+
+async function withDisconnectedTransaction(fn) {
+  const pool = postgres({ ...options, fetch_types: false })
+  let finish
+    , ready
+  const gate = new Promise((resolve, reject) => finish = error => error ? reject(error) : resolve())
+  const connected = new Promise(resolve => ready = resolve)
+  const failed = pool.begin(async sql => {
+    const [{ pid }] = await sql`select pg_backend_pid() as pid`
+    ready({ disconnected: sql, pid })
+    await gate
+  }).catch(x => x)
+
+  try {
+    const { disconnected, pid } = await Promise.race([connected, failed.then(error => { throw error })])
+    await sql`select pg_terminate_backend(${ pid }::int)`
+    await failed
+    return await fn({ sql: pool, disconnected, finish })
+  } finally {
+    finish()
+    await new Promise(resolve => setImmediate(resolve))
+    await pool.end({ timeout: 0 })
+  }
+}
+
 t('Custom socket', {}, async() => {
   let result
   const sql = postgres({


AlephNotation:
I've also recently gotten bitten by this one

batalyse-gmbh:
Thanks for this PR. We tested it together with #1209: the two fix every crash case we could reproduce. We found one case that is still broken, and a one-line fix for it.

### What we tested

- postgres 3.4.9, Node 26, PostgreSQL 18.4 and 18.6 (same results on both)
- `max: 1`, so the same connection is always reused

### On 3.4.9 without the PRs

| Scenario | What happens |
|---|---|
| `transaction_timeout` or `pg_terminate_backend` during a query in `sql.begin()` | Crash: the `nextWrite` TypeError from #1133. If you catch it, the connection never comes back: every reconnect ends in `CONNECT_TIMEOUT`. |
| `idle_in_transaction_session_timeout`, then another `begin()` reuses the connection | The dead transaction's INSERT and COMMIT run inside the other transaction. That transaction throws, but its inserts stay committed (#1216). |
| Any kill during a query | The next query fails with the old backend's 57P01 / 25P04 error. |

A plain null check in `nextWrite` (as in #1168) is not enough. It turns the crash into the same endless `CONNECT_TIMEOUT`.

### With #1209 + #1215

All three are fixed. ✅

### Still broken: a connection reset inside `begin()`

**Symptom:** after an ECONNRESET (for example a failover), one query works. After that the connection hangs forever.

**Cause:**

1. On a reset, the socket emits `'error'` first and `'close'` a moment later. Promise callbacks run in between.
2. In that gap, `begin()` sends its `rollback`. The socket is still set and `closedError` is not set yet, so the rollback is added to `sent`.
3. `closed(true)` only clears `sent` when `!hadError`, so the rollback stays there.
4. After reconnecting, the connection treats that old rollback as in flight and waits for a reply that never comes.

**Repro:**

```js
import net from "node:net";

// A TCP proxy in front of Postgres, so we can reset the socket like a failover would.
const pairs = [];
const proxy = net.createServer((client) => {
  const upstream = net.connect(5432, "127.0.0.1");
  client.pipe(upstream).pipe(client);
  client.on("error", () => {});
  upstream.on("error", () => {});
  pairs.push([client, upstream]);
}).listen(0, "127.0.0.1");
await new Promise((r) => proxy.once("listening", r));

const sql = postgres({ host: "127.0.0.1", port: proxy.address().port, max: 1 /* , user, password */ });

await sql.begin(async (tx) => {
  setTimeout(() => pairs.forEach(([c, u]) => (u.destroy(), c.resetAndDestroy())), 500);
  await tx`select pg_sleep(3)`;
}).catch((e) => console.log("begin rejected:", e.code));

const ask = (tag) => sql`select ${tag}::text as tag`.then(([r]) => console.log(tag, "->", r.tag));
await ask("Q1");
await Promise.race([ask("Q2"), new Promise((r) => setTimeout(() => r(console.log("Q2 still pending after 3s")), 3000))]);
process.exit(0);
```

Output on 3.4.9, and on 3.4.9 + #1209 + #1215:

```
begin rejected: CONNECTION_CLOSED
Q1 -> Q1
Q2 still pending after 3s
```

Expected: `Q2 -> Q2`.

### Fix

Always clear `sent` when the socket closes:

```diff
-    !hadError && (query || sent.length) && error(Errors.connection('CONNECTION_CLOSED', options, socket))
+    if (query || sent.length)
+      error(Errors.connection('CONNECTION_CLOSED', options, socket))
```

It has to be an `if`. The code has no semicolons, so a line starting with `(` would join onto the `return reconnect()` above it.

We tested it on top of #1209 + #1215:
- the repro above passes;
- a plain query outside a transaction still gets `ECONNRESET`;
- all the scenarios in the table still pass.

*Written by Claude Opus*

cat-a-guerra:
We independently found and patched the same two issues in production (3.4.9): the stale `chunk`/`nextWriteTimer` that leaves reconnects ending in `CONNECT_TIMEOUT`, and the stale `query`/`errorResponse` that replays the old backend's 57P01 onto the next query. Our patch makes the same two resets, so +1 to this approach. We also saw `sql.end()` never resolve while the connection was in that stale state.

One path this PR doesn't reach: `closed()` returns early via `if (initial) return reconnect()` before the new reset lines run. So a connection killed during startup (the `fetch_types` query) still replays its 57P01 onto the first query, plus an unhandled rejection. Repro and a one-line fix in #1223. We confirmed this PR alone doesn't change the output there.

Issue #970, opened 2024-10-27
Deadlocked/hung pool when connected to Postgres through Supavisor


nikhilro:
Hey @porsager, thanks for the package. 

Creating this issue mostly in hopes that someone else has ran into this issue.

We use Supabase for our Postgres DB and connect through their pooling service Supavisor in transaction mode; Supavisor is an alternative to pgBouncer ([link](https://github.com/supabase/supavisor)). On postgres.js side, we set `prepare: false`.

What we're seeing is that the postgres.js pooler will start swallowing queries at some point. Imagine a function like:
```
async function userGet(id: string) {
  console.log("trying to get the user");
  const users = await db<User[]>`SELECT * FROM user WHERE id=${id}`;
  console.log("got the user!");
  return users;
}
```

After a while, this will print "trying to get the user" but not "got the user!". Does this ring any bells?

nikhilro:
I'm more confident that is an issue with Supavisor under load. A thing that would be helpful is postgres.js statement limit. 

I know we have:
```
   connection: {
     statement_timeout: 1000 * 60 * 0.5, // 30 seconds, pg expects milliseconds
   },
```

But, that is not useful when connecting to transaction mode poolers. They share their setting across the clients.

Is there an easy way to "abort" the query if it's taking too long?

lllleonnnn:
Please forgive not answering your exact question but we have run into very similar issues and it boiled down to our queries being blocking & locking.  We've used this query (modify as needed) to run down issues in our Supabase PG instance:
```sql
SELECT blocked_locks.pid AS blocked_pid,
 blocked_activity.usename AS blocked_user,
 blocking_locks.pid AS blocking_pid,
         blocking_activity.usename AS blocking_user, 
         blocked_activity.query    AS blocked_statement,
         blocking_activity.query   AS current_statement_in_blocking_process
   FROM  pg_catalog.pg_locks         blocked_locks 
    JOIN pg_catalog.pg_stat_activity blocked_activity  ON blocked_activity.pid = blocked_locks.pid
    JOIN pg_catalog.pg_locks         blocking_locks 
        ON blocking_locks.locktype = blocked_locks.locktype 
        AND blocking_locks.DATABASE IS NOT DISTINCT FROM blocked_locks.DATABASE
        AND blocking_locks.relation IS NOT DISTINCT FROM blocked_locks.relation
        AND blocking_locks.page IS NOT DISTINCT FROM blocked_locks.page
[...]

PR #1218, opened 2026-09-09
Fix sql.begin() sending BEGIN on an unreserved connection

### LINKED ISSUE #1189: BEGIN can reach PostgreSQL without reserving the transaction connection
## Summary

In `postgres@3.4.9`, a transaction's `BEGIN` can be written to the socket while its `onexecute(connection)` reservation hook is skipped.

The affected ordering is currently present on `master` in both Node entry points:

```js
return write(toBuffer(q))
  && !q.describeFirst
  && !q.cursorFn
  && sent.length < max_pipeline
  && (!q.options.onexecute || q.options.onexecute(connection))
```

At the pipeline boundary, adding `BEGIN` to `sent` makes `sent.length === max_pipeline`, so short-circuit evaluation skips `onexecute` even though the protocol bytes were accepted. `sql.begin()` then has no reserved connection, while PostgreSQL has already entered a transaction.

This is related to #823 and the closed, unmerged #1159. Those identify the intermittent `UNSAFE_TRANSACTION`/pipeline symptom. The deterministic case below additionally demonstrates the server-side isolation impact when `max: 1`: the connection returns to ordinary use as `idle in transaction`, and the next unrelated query runs on the same backend.

## Environment

- `postgres@3.4.9` from the npm tarball, unmodified
- Node.js 24 (the ESM case was included in the private disclosure; the equivalent CJS case was subsequently reproduced as well)
- PostgreSQL 17
- Reproduced repeatedly at the exact pipeline boundary

## Reproduction scope

The reproduction below deterministically isolates the exact-pipeline-boundary trigger. It still performs the real socket write; forcing the wrapper to return `true` only removes socket backpressure from this case.

A separate trigger exists when the native `socket.write()` returns `false` after accepting the bytes: the same short-circuit expression skips `onexecute` before reaching the pipeline check. That backpressure path is supported by source inspection but is not independently reproduced by the script below.

## Minimal reproduction

Run against a disposable database:

```sh
DATABASE_URL=postgres://... node repro-transaction-reservation.mjs
```

```js
import assert from 'node:assert/strict'
import net from 'node:net'

const databaseUrl = process.env.DATABASE_URL
if (!databaseUrl) throw new Error('Set DATABASE_URL to a disposable PostgreSQL database')
const endpoint = new URL(databaseUrl)

async function connectSocket() {
  const socket = net.createConnection({
    host: endpoint.hostname,
    port: Number(endpoint.port || 5432),
  })
  await new Promise((resolve, reject) => {
    socket.once('connect', resolve)
    socket.once('error', reject)
  })

  // Preserve the real write, but make this boundary deterministic.
  const nativeWrite = socket.write
  socket.write = function acceptedWrite(chunk, ...args) {
    Reflect.apply(nativeWrite, this, [chunk, ...args])
    return true
  }
  return socket
}

const sql = postgres(databaseUrl, {
  fetch_types: false,
  max: 1,
  max_pipeline: 1,
  prepare: false,
  socket: connectSocket,
})
const observer = postgres(databaseUrl, { max: 1, prepare: false })

// Stock 3.4.9 can emit a secondary rejection after the primary begin rejection.
const onUnhandled = () => {}
process.on('unhandledRejection', onUnhandled)

try {
  await sql`select 1`

  const active = Promise.resolve(sql`
    select pg_backend_pid()::integer as pid
    from pg_sleep(0.25)
  `.execute())

  const transaction = Promise.resolve(sql.begin(async () => true))
  const transactionObserved = transaction.then(
    () => ({ status: 'fulfilled', error: null }),
    (error) => ({ status: 'rejected', error }),
  )

  const [{ pid }] = await active
  const transactionResult = await Promise.race([
    transactionObserved,
    new Promise((resolve) => setTimeout(
      () => resolve({ status: 'still_pending', error: null }),
      1500,
    )),
  ])

  const [activity] = await observer.unsafe(
    'select state from pg_stat_activity where pid = $1::integer',
    [pid],
  )

  const [{ reusedPid }] = await sql`
    select pg_backend_pid()::integer as "reusedPid"
  `

  console.log({
    transactionStatus: transactionResult.status,
    transactionError: transactionResult.error?.message ?? null,
    backendPid: pid,
    backendState: activity?.state,
    nextQueryBackendPid: reusedPid,
  })

  assert.equal(transactionResult.status, 'rejected')
  assert.match(transactionResult.error?.message ?? '', /undefined/)
  assert.equal(activity?.state, 'idle in transaction')
  assert.equal(reusedPid, pid)
} finally {
  await Promise.allSettled([
    sql.end({ timeout: 0 }),
    observer.end({ timeout: 1 }),
  ])
  await new Promise((resolve) => setImmediate(resolve))
  process.off('unhandledRejection', onUnhandled)
}
```

Observed repeatedly:

```text
{
  transactionStatus: 'rejected',
  transactionError: "Cannot set properties of undefined (setting 'onclose')",
  backendPid: <pid>,
  backendState: 'idle in transaction',
  nextQueryBackendPid: <same pid>
}
```

## Impact

The next logical caller can execute on a physical connection that is still inside the orphaned transaction. A later `COMMIT`, `ROLLBACK`, disconnect, or transaction timeout can therefore commit or roll back work belonging to a different caller.

This is a data-integrity and cross-request isolation issue, not only a failed transaction callback.

## Expected behavior

For a query carrying `options.onexecute`:

- once its protocol bytes have been accepted by `write()` without throwing, reserve the connection exactly once;
- do not make reservation conditional on `sent.length < max_pipeline`;
- treat `socket.write() === false` as accepted with backpressure, not as unsent;
- after reservation, do not offer the connection to ordinary pipeline work;
- preserve the existing describe-first and cursor behavior;
- cover both ESM and CJS at the exact pipeline boundary and under backpressure.

I can prepare a narrowly scoped PR and regression tests if that would help.

v0idpwn:
sql.begin() could send BEGIN to the server without reserving the
connection.

Depending on pool size, this errored with UNSAFE_TRANSACTION or
TypeError.  The transaction stayed open on the server either way, and
the connection went back to the pool, so unrelated queries then ran
inside it.

With max_pipeline: 0 this error happened consistently.  With the default
max_pipeline of 100 the same thing happened only when BEGIN was the
query that filled a connection's pipeline, which is why the failure was
intermittent.

The issue is in execute(): after writing the query it returns an &&
chain that states whether the connection can accept more queries, and
one of its terms is sent.length < max_pipeline.  The onexecute hook,
which is how sql.begin() reserves the connection, was the last term of
that chain.  Whenever the capacity term was false the chain
short-circuited and the hook never ran, even though BEGIN had already
been written.  With max_pipeline: 0 that term is false for every query,
so the hook never ran at all.

Move the hook ahead of the capacity terms so it runs whenever the query
was written.  Reserving the connection is a consequence of having sent
BEGIN, unrelated to the query limit in the connection.

Add tests for BEGIN at the pipeline boundary and with max_pipeline: 0.

Fixes https://github.com/porsager/postgres/issues/1189 https://github.com/porsager/postgres/issues/1210

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -170,11 +170,16 @@ function Connection(options, queues = {}, { onopen = noop, onend = noop, onclose
         : (query = q, query.active = true)
 
       build(q)
-      return write(toBuffer(q))
+      const written = write(toBuffer(q))
+      // Run the hook whenever the bytes were written, even if the pipeline is
+      // full or socket.write() reported backpressure. sql.begin() relies on it
+      // to reserve the connection, and its falsy return keeps the connection
+      // out of the busy queue until BEGIN completes.
+      return (!q.options.onexecute || q.options.onexecute(connection))
+        && written
         && !q.describeFirst
         && !q.cursorFn
         && sent.length < max_pipeline
-        && (!q.options.onexecute || q.options.onexecute(connection))
     } catch (error) {
       sent.length === 0 && write(Sync)
       errored(error)
--- a/tests/index.js
+++ b/tests/index.js
@@ -300,6 +300,21 @@ t('Many transactions at beginning of connection', async() => {
   return [100, xs.length]
 })
 
+t('Transaction at pipeline boundary is reserved', async() => {
+  const sql = postgres({ ...options, max: 2, max_pipeline: 1, fetch_types: false })
+  await Promise.all([sql`select 1`, sql`select 1`])
+  const inflight = [sql`select pg_sleep(0.1)`.execute(), sql`select pg_sleep(0.1)`.execute()]
+  const x = await sql.begin(sql => sql`select 1 as x`).then(x => x[0].x, x => x.code)
+  await Promise.all(inflight)
+  return [1, x, await sql.end()]
+})
+
+t('Transaction is reserved with pipelining disabled', async() => {
+  const sql = postgres({ ...options, max: 2, max_pipeline: 0, fetch_types: false })
+  const x = await sql.begin(sql => sql`select 1 as x`).then(x => x[0].x, x => x.code)
+  return [1, x, await sql.end()]
+})
+
 t('Transactions array', async() => {
   await sql`create table test (a int)`
 
--- a/tests/index.js
+++ b/tests/index.js
@@ -315,6 +315,18 @@ t('Transaction is reserved with pipelining disabled', async() => {
   return [1, x, await sql.end()]
 })
 
+t('Query issued while BEGIN is in flight does not join the transaction', async() => {
+  const sql = postgres({ ...options, max: 1, fetch_types: false })
+  await sql`create table test (a int)`
+  const tx = sql.begin(async sql => {
+    await sql`select 1`
+    throw new Error('rollback')
+  }).catch(() => {})
+  const insert = sql`insert into test values (1)`
+  await Promise.all([tx, insert])
+  return [1, (await sql`select count(*)::int as n from test`)[0].n, await sql`drop table test`, await sql.end()]
+})
+
 t('Transactions array', async() => {
   await sql`create table test (a int)`
 


Ruckus000:
A production data point for this fix, and for #970 / #1210.

- **Setup:** Next.js on Vercel, Supabase Supavisor in transaction mode (port 6543), postgres.js 3.4.8 via drizzle-orm, `prepare: false`, `max: 3`.
- **Symptom:** a page that issued 8 concurrent queries hung until the platform's 300s timeout, on every load. The query that postgres.js had pipelined onto a busy connection never resolved, and PostgreSQL showed its backend `active` / wait event `ClientRead`.
- **Confirmation:** raising `max` to 10, so nothing needed to pipeline, made the page load; that was the confirming test. We then shipped `max_pipeline: 0` with `max: 3`.
- **Patch:** we had to patch `execute()` to run `onexecute` unconditionally, because otherwise every `sql.begin` throws (#1210).

We run that patch in production. Its semantics match this PR: call the hook whenever the bytes were written, and keep a falsy return while it's present. In our local tests it passes concurrent transactions, savepoints and rollback at `max_pipeline: 0`, and keeps the transaction's connection out of ordinary pipelining at the default and at `1`. (See my comment on #1211 for why that last part matters.)

